Issue #2, opened 2021-12-30
Create environment.yml


AndyMcAliley:
Allow conda environment to be replicated via `conda create -f environment.yml`.

A few expected dependencies:
- pytorch (Specify version, make sure gpu is enabled)
- numpy
- pandas
- matplotlib
- R
- sbtools
- secret
- dplyr

AndyMcAliley:
TBD if we'll actually need to use `R` in any way.

PR #6, opened 2022-01-08
Fetch MNTOHA data from ScienceBase using Snakemake

### LINKED ISSUE #1: Create Snakefile
Since most of this repository will be in Python, we'll use [Snakemake](https://snakemake.readthedocs.io/en/stable/) to manage the workflow. An example of using Snakemake in our branch can be found in [river-dl](https://github.com/USGS-R/river-dl). Some downloads may be easier in R (especially from ScienceBase, using `sbtools`), so we want to make sure that Snakemake is capable of handling calls to R gracefully.

- Snakemake can execute shell commands, so at the least we can call `Rscript`.
- Snakemake can also [call R scripts](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#r-and-r-markdown). An S4 object named snakemake is created. Inputs and outputs can be accessed using `snakemake@input[["myfile"]]`.

### LINKED ISSUE #4: Download MNTOHA data to get started
Download MNTOHA data (I think all from ScienceBase):

- Temperature observations
- Lake attributes
  - Lake metadata
  - (?) GLM model configurations
- NLDAS drivers

AndyMcAliley:
This is a first pass at `1_fetch`, using MNTOHA since that download is straightforward. Eventually we'll use more lakes than MNTOHA. This uses Snakemake to manage the pipeline. Closes #1 and closes #4. #2 is addressed except for ensuring that pytorch is gpu-enabled.

# `Snakemake`

This is my first time using `Snakemake`, so I'm probably not using it optimally. Still, here's a walkthrough of how it's set up. I set this up while consulting the [docs](https://snakemake.readthedocs.io/en/stable/index.html), [drb-estuary-salinity-ml](https://github.com/USGS-R/drb-estuary-salinity-ml), and [river-dl](https://github.com/USGS-R/river-dl).

The Snakefile contains all the rules for what needs to be made. The first rule is always the default rule that gets made if you call `snakemake` from the command line without specifying any particular rule. So, generally, that rule is the "all" rule. Here, it's called `fetch_all`. It ensures that every file we want is, in fact, downloaded in the proper place. It does that by requiring all those files as inputs. This rule alone has no outputs because it relies on other rules to actually do the downloading - it just lists all the files we want as dependencies. It also has one other input: `1_fetch/in/pull_date.txt`. This file just lists a date. It's a dummy file that can be altered when we want to trigger a fresh download.

The way `fetch_all` requires all the files we want is by referencing the config dictionary, which is loaded from `1_fetch/fetch_config.yaml`. There, you'll find every individual file listed, plus ScienceBase IDs.

Running `snakemake -c1 -p fetch_all` specifies this rule, which downloads everything.

The other rules are "[generalized](https://snakemake.readthedocs.io/en/stable/tutorial/basics.html#step-2-generalizing-the-read-mapping-rule)" rules for how to download files with file paths that match a specific pattern. For instance, say we want to download `1_fetch/out/metadata/lake_metadata.csv`. `Snakemake` notices that this file path matches the pattern `1_fetch/out/metadata/*`, which is the pattern specified in the output of the `fetch_metadata` rule. So, `Snakemake` uses the `fetch_metadata` rule for that file. Likewise, The `fetch_obs` rule is for any files downloaded to `1_fetch/out/obs/`, and the `fetch_driver` rule is for any files downloaded to `1_fetch/out/driver/`.

The `shell` directive is followed by the shell command to run. That command is expected to produce the output files in the `output` directive. Here, the `shell` commands for the three rules all call the same python script, `1_fetch/src/sb_fetch.py`, but with different arguments. The first argument is the ScienceBase ID, and the second argument is the file path. The ScienceBase ID is drawn from the `Snakemake` config file, found at `1_fetch/fetch_config.yaml`. Each rule has its own ScienceBase ID associated with it. 

This aspect of the pipeline is a bit brittle. Since the download path is used to determine which rule to use, and each rule is associated with one SB ID, that means that every SB ID has to have its own subdirectory in `1_fetch/out/`. Not a huge deal, but I suspect there's a more robust way. It may not matter, though, since we'll eventually move away from MNTOHA and probably SB downloads. I do want to get my head around `Snakemake` best practices, though.

The good news is that this method makes `Snakemake` aware of every file to download. This way, files can be downloaded in parallel. If one file is missing or the download failed partway through, calling `Snakemake` again will cause it to pick up where it left off, without needing to re-download everything from that SB ID.


1_fetch/fetch_config.yaml
```diff
@@ -0,0 +1,68 @@
+sb_fetch.py:
```
jsadler2, line 1: I would consider restructuring this config file with another level, the category. (`metadata`, `obs`,  `drivers`). 

See example here: https://github.com/jsadler2/lake-temperature-lstm-static/blob/snakefile-shorter/1_fetch/fetch_config.yaml

1_fetch/fetch_config.yaml
```diff
@@ -0,0 +1,68 @@
...
+      - 1_fetch/out/metadata/lake_metadata.csv
+
+  obs_id: 5e5d0b68e4b01d50924f2b32
+  obs_files:
+    - 1_fetch/out/obs/temperature_observations.zip
+
+  driver_id: 5e5d0b96e4b01d50924f2b34
+  driver_files:
+    - 1_fetch/out/driver/04_inputs.xml
+    - 1_fetch/out/driver/clarity_01_N48.00-49.50_W89.50-97.25.zip
```
jsadler2, line 14: If you use `expand` and separate out the "driver types" (`clarity`, `ice_flags`, `inputs`) you can let snakemake come up with all of the combinations of these.

* `expand` example: https://github.com/jsadler2/lake-temperature-lstm-static/blob/ef32613bf46a67df06a80195472ac4518718be28/Snakefile#L13

* breaking out "driver types" example: https://github.com/jsadler2/lake-temperature-lstm-static/blob/ef32613bf46a67df06a80195472ac4518718be28/1_fetch/fetch_config.yaml#L18

1_fetch/fetch_config.yaml
jsadler2: This allows you to access the `sb_id` more easily: https://github.com/jsadler2/lake-temperature-lstm-static/blob/ef32613bf46a67df06a80195472ac4518718be28/Snakefile#L24

README.md
```diff
@@ -1,2 +1,38 @@
...
+```
+conda update -n base conda
+conda env create -f environment.yaml
+conda activate ltls
+```
+
+Note: if [Mamba](https://github.com/mamba-org/mamba) is available, the environment may be created significantly faster by running
+
+```
+mamba update -n base mamba
```
jsadler2, line 18: I didn't know about `mamba`. That looks helpful!

Snakefile
```diff
@@ -0,0 +1,37 @@
...
+
+rule fetch_all:
+    input:
+        "1_fetch/in/pull_date.txt",
+        expand("{file}", file=config["sb_fetch.py"]["metadata_files"]),
+        expand("{file}", file=config["sb_fetch.py"]["obs_files"]),
+        expand("{file}", file=config["sb_fetch.py"]["driver_files"])
+
+# Lake information for 881 MNTOHA lakes
+rule fetch_metadata:
```
jsadler2, line 12: with restructuring the config file, you could use just one rule (see https://github.com/jsadler2/lake-temperature-lstm-static/blob/ef32613bf46a67df06a80195472ac4518718be28/Snakefile#L18)

1_fetch/src/sb_fetch.py
```diff
@@ -0,0 +1,91 @@
...
+            filename = None
+            item_str = item_id
+        else:
+            item_str = filename
+    sb = SbSession()
+    print(f'Downloading {item_str} to {destination_dir}')
+    response = sb_get(sb, item_id, filename, destination_dir=destination_dir)
+
+
+def parse_args():
```
jsadler2, line 71: If you use `script` instead of `shell`, Snakemake does all of the command line argument parsing for you:

* see simplified snakefile: https://github.com/jsadler2/lake-temperature-lstm-static/blob/62407e627e9feb079700d9c816c0cb13fc7d8b94/Snakefile#L32

* and simplified .py file: https://github.com/jsadler2/lake-temperature-lstm-static/blob/62407e627e9feb079700d9c816c0cb13fc7d8b94/1_fetch/src/sb_fetch.py#L69

Snakefile
```diff
@@ -0,0 +1,37 @@
+configfile: "1_fetch/fetch_config.yaml"
```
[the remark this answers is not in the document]
jsadler2: if you do this, I would read the Snakemake docs about modularization (https://snakemake.readthedocs.io/en/stable/snakefiles/modularization.html). 

I think it would probably be simplest while also being sufficient to use the `include` funcationality where each phase's snakefile includes the relevant previous phases' snakefiles.

jsadler2 (commented):
Hey @AndyMcAliley, exciting to see this!

As we discussed on the call yesterday, I have a few suggestions and alternative ways you might think about using Snakemake for your pipeline.

Here are three different alternatives that you may find helpful.
* [restructuring config file and using expand to combine the suffixes and "driver types"](https://github.com/jsadler2/lake-temperature-lstm-static/tree/snakefile-shorter)
* [above + using input function to automatically populate the other input files](https://github.com/jsadler2/lake-temperature-lstm-static/blob/snakefile-even-shorter/Snakefile)
* [above + using `script` instead of `shell`](https://github.com/jsadler2/lake-temperature-lstm-static/tree/snakefile-script)

I would most suggest the four following adjustments:
1. use expand for the driver files with suffixes
2. add one more level to config so you can use just one rule
3. rename your snakemake file to 1_fetch.smk
4. use `script` instead of `shell` to save yourself some code/mental space

Snakefile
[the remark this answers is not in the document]
AndyMcAliley: Done! I also added a new `Snakefile` which is the main file containing the rule `all`.

1_fetch/fetch_config.yaml
AndyMcAliley: Done! This structure does make things easier.

1_fetch/fetch_config.yaml
AndyMcAliley: I like how short the config file is with this change. However, rather than let Snakemake create the combinations, I moved that logic into the `get_input_files` function (now renamed as `get_mntoha_input_files`). That feels a bit easier to read and maintain to me?

README.md
AndyMcAliley: I just started using `mamba`. It was the `Snakemake` docs that pointed me to it. It is quick, especially when it comes to resolving package conflicts, and it's been easy to use as a drop-in replacement for `conda` in most circumstances.

Snakefile
AndyMcAliley: Done! This is nice.

1_fetch/src/sb_fetch.py
AndyMcAliley: Done! Using `script` does make things cleaner.

AndyMcAliley:
The workflow should be general enough to accommodate future data downloads from other Sciencebase items besides MNTOHA, or from sources other than Sciencebase. I renamed a number of things from "sb" to "mntoha" to try to make room for those future changes. In `1_fetch/out`, I also added `_mntoha` as a suffix to each folder in anticipation of pulling other data later. That suffix can help Snakemake figure out how to do the download.

1_fetch.smk
```diff
@@ -0,0 +1,43 @@
...
+    :param wildcards: Snakemake wildcards (unused)
+    :returns: list of filepaths to be downloaded
+
+    '''
+    out_dir = "1_fetch/out"
+    categories = config["mntoha"].keys()
+    input_files = []
+    # include files under files key in each category
+    for category in categories:
+        for filename in config["mntoha"][category]["files"]:
```
hcorson-dosch, line 19: gosh I love dictionaries.

1_fetch.smk
```diff
@@ -0,0 +1,43 @@
...
+            input_files.append(os.path.join(out_dir, category + "_mntoha", filename))
+    # include driver files: 3 types x 13 suffixes = 39 files
+    for driver_type in config["mntoha"]["drivers"]["driver_types"]:
+        for suffix in config["mntoha"]["drivers"]["suffixes"]:
+            input_files.append(os.path.join(out_dir, "drivers_mntoha", f"{driver_type}_{suffix}.zip"))
```
hcorson-dosch, line 24: And f-strings! So much cleaner than `sprintf()` in R!

1_fetch.smk
```diff
@@ -0,0 +1,43 @@
+import os
+
+configfile: "1_fetch/fetch_config.yaml"
+
+
+def get_mntoha_input_files(wildcards):
+    '''List all MNTOHA files to download
+    Includes metadata, temperature observations, and drivers
+
+    :param wildcards: Snakemake wildcards (unused)
```
hcorson-dosch, line 10: Isn't `wildcards` used in line 40?

1_fetch.smk
```diff
@@ -0,0 +1,43 @@
...
+        get_mntoha_input_files
+
+
+rule fetch_mntoha_data_file:
+    input:
+        "1_fetch/in/pull_date.txt"
+    output:
+        "1_fetch/out/{file_category}_mntoha/{file}"
+    params:
+        sb_id = lambda wildcards: config["mntoha"][wildcards.file_category]["sb_id"]
```
hcorson-dosch, line 40: Just trying to wrap my head around what `snakemake` is doing here. So this works b/c `snakemake` automatically uses the `mntoha` dictionary from `1_fetch/fetch_config.yml` as `wildcards` and then the keys from that dictionary as `file_category`?

1_fetch.smk
```diff
@@ -0,0 +1,43 @@
...
+    input:
+        "1_fetch/in/pull_date.txt",
+        get_mntoha_input_files
+
+
+rule fetch_mntoha_data_file:
+    input:
+        "1_fetch/in/pull_date.txt"
+    output:
+        "1_fetch/out/{file_category}_mntoha/{file}"
```
hcorson-dosch, line 38: So `snakemake` knows this rule applies to all of the files returned by `get_mntoha_input_files` because they all match this  pattern, where {file_category} and {file} aren't actually known to `snakemake`, but it can infer them based on this pattern matching?

1_fetch.smk
AndyMcAliley: Line 40 is outside the scope of `get_mntoha_input_files`, so that `wildcards` is a different variable. We could rename one for clarity.

1_fetch.smk
AndyMcAliley: Okay, here's my understanding. So, if the rule `fetch_mntoha_data_file` is being called, that means that `snakemake` has matched a requested file path to the pattern defined here in `output`. [`wildcards`](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#wildcards) contains both of the wildcards that were named in the pattern defined in `output`: `file_category` and `file`. If a function is called from within a `snakemake` rule, the first argument passed to that function is `wildcards`. That goes for both named functions (e.g. `get_mntoha_input_files`) and `lambda` functions like the one in line 40. That's why `wildcards` is named the same in lines 6 and 40 - it's that argument - even though in line 6, we know it's the `wildcards` from the `fetch_all` rule, while in line 40, it's the `wildcards` from the `fetch_mntoha_data_file` rule. From there, like you said, `snakemake` gets the `sb_id` by indexing the `config` that's loaded from `1_fetch/fetch_config.yaml`.

AndyMcAliley: Alternatively, we could have defined a function, like:
```
def get_sb_id(wildcards):
    file_category = wildcards.file_category
    return config["mntoha"][file_category]["sb_id"]

rule fetch_mntoha_data_file:
    input:
        "1_fetch/in/pull_date.txt"
    output:
        "1_fetch/out/{file_category}_mntoha/{file}"
    params:
        sb_id = get_sb_id
    script:
        "1_fetch/src/sb_fetch.py"


```

1_fetch.smk
AndyMcAliley: Right! Whenever a file path matches `1_fetch/out/*_mntoha/*`, `snakemake` calls this rule to make the file, where the wildcard `file_category` is the first `*` and `file` is the second `*`.

README.md
```diff
@@ -1,2 +1,38 @@
...
+
+Predict lake temperatures at multiple depths using an LSTM trained to make use of static lake attributes.
+
+## Installing
+
+Create a new `conda` environment by running
+
+```
+conda update -n base conda
+conda env create -f environment.yaml
```
hcorson-dosch, line 11: So nice to have the environment captured in a yaml file!

README.md
```diff
@@ -1,2 +1,38 @@
...
+
+## Downloading input data
+
+For now, MNTOHA data are being used.
+A more complete set of lakes will be incorporated in the future, but MNTOHA lake data are readily available to download from ScienceBase.
+To download temperature observations, meteorological drivers, and metadata, run
+
+`snakemake -c1 -p fetch_all`
+
+Replace the `1` in `-c1` with the number of cores to execute jobs in parallel.
```
hcorson-dosch, line 31: So running jobs in parallel is that easy?! `snakemake` does it automatically if the job is allocated multiple cores?

Snakefile
```diff
@@ -0,0 +1,8 @@
+include: "1_fetch.smk"
```
hcorson-dosch, line 1: So here's where rules associated with multiple phases of the code would be pulled together in to a master file, much like `_targets.R`?

1_fetch.smk
hcorson-dosch: Thanks, Andy - that makes complete sense now.

1_fetch.smk
AndyMcAliley: That's one thing about `snakemake` that I'm still wrapping my head around. It means that we should be careful about naming conventions, since file path matching is how `snakemake` determines how to use one rule to make multiple files.

1_fetch.smk
jsadler2: I agree with your explanation, @AndyMcAliley.

I'll add a concrete example, in case that is helpful. 

- Snakemake always works from what you are asking it to make. In `fetch_all`, one of the things we are asking it to make is `1_fetch/out/drivers/input_cells.png` (one file in the list of many files that the `get_mntoha_input_files` function returns).
- Snakemake finds the rule that makes `1_fetch/out/drivers/input_cells.png`. This is `fetch_mntoha_data_file` rule. 
- The `wildcards` object gets built by parsing the filename according to the wildcards defined in the `output` string of the rule (`"1_fetch/out/{file_category}_mntoha/{file}"`). 
- So in this case `wildcards.file_type = "drivers"` and `wildcards.file = "input_cells.png"`. 
- Then, the `params` are defined. In this case, the one-line `lambda` function will return `config["mntoha"]["drivers"]["sb_id"]` which is `"5e5d0b96e4b01d50924f2b34"` from the config file. 
- That `sb_id` is then accessed via the `params` object in the `1_fetch/src/sb_fetch.py` file since you are using the `script` functionality.

README.md
AndyMcAliley: I haven't tried this on HPC resources, but it worked on my laptop for the download!

Snakefile
AndyMcAliley: Exactly. I believe if you call `snakemake` from the command line without specifying a file, it defaults to `Snakefile`.

1_fetch.smk
hcorson-dosch: That is helpful -- thanks, Jeff!

1_fetch/src/sb_fetch.py
```diff
@@ -0,0 +1,70 @@
...
+            item_str = item_id
+        else:
+            item_str = filename
+    sb = SbSession()
+    print(f'Downloading {item_str} to {destination_dir}')
+    response = sb_get(sb, item_id, filename, destination_dir=destination_dir)
+
+
+if __name__ == '__main__':
+    main(snakemake.params['sb_id'], snakemake.output[0])
```
hcorson-dosch, line 70: Hmm I'm a bit confused -- if no files were listed under a given category (in this example, `metadata`) in the `mntoha` dictionary defined in `1_fetch/fetch_config.yaml`, like so:
```
mntoha:

  metadata:
      sb_id: 5e5c1c1ce4b01d50924f27e7
      files:

  obs:
      sb_id: 5e5d0b68e4b01d50924f2b32
      files:
          - temperature_observations.zip
```
Then wouldn't no files (nor even a directory) be added for that category by the `get_mntoha_input_files()` function, which would then mean that the `fetch_mntoha_data_file` rule wouldn't identify `metadata` as a `file_category`? And therefore the `1_fetch/src/sb_fetch.py` script wouldn't be called for that category?

Just trying to think through when the `filepath = None` default argument would be used, and I'm not sure how it would work given this `snakemake` setup. But I may be missing something.

1_fetch/src/sb_fetch.py
```diff
@@ -0,0 +1,70 @@
...
+    if filepath is None:
+        filename = None
+        destination_dir='.'
+        item_str = item_id
+    else:
+        filename = os.path.basename(filepath)
+        destination_dir = os.path.dirname(filepath)
+        if not os.path.exists(destination_dir):
+            os.makedirs(destination_dir)
+        if filename == '':
```
hcorson-dosch, line 59: Similarly to my comment on line 70 - when would this be the case, given the current setup? if you provided this, with just a space or nothing after the hash?
```
mntoha:

  metadata:
      sb_id: 5e5c1c1ce4b01d50924f27e7
      files:
         -  
  obs:
      sb_id: 5e5d0b68e4b01d50924f2b32
      files:
          - temperature_observations.zip
```
And if that is simply a space (or an empty string) for the `['metadata']['files']` value in the `mntoha` dictionary, then does `snakemake` match `''`  for `{file}` in the wildcard using the `"1_fetch/out/{file_category}_mntoha/{file}"` pattern?

1_fetch/src/sb_fetch.py
AndyMcAliley: You're right - the way it's set up now, `filepath=None` wouldn't be used at any point, and that code in `sb_fetch` doesn't make sense. I wrote that code when I was using `snakemake` in a different way, using `shell` instead of `script`. I can remove the logic for `filepath=None` from `main`.

1_fetch/src/sb_fetch.py
AndyMcAliley: Right, as above, that logic won't get triggered by `snakemake` either.

AndyMcAliley: If we wanted to download all the files in one SB item... we could rewrite `main` in `sb_fetch.py`, and probably include a boolean in Snakemake.params that signifies whether we want the single files or all files in the SB item. Do you think we'll want that functionality at some point?

AndyMcAliley: (I tried a few iterations of None/empty string for `['metadata']['files']` and got errors)

hcorson-dosch: I think that leaving as is (having to specify the files) is fine for now, given the current use case, which is the mntoha data. If we switch to a new data sources, that might either 1) motivate adding the option to download all files from a given SB id or 2) require changes to this framework anyhow (e.g., a total hypothetical, but maybe we decide to store data on caldera instead and pull with globus sdk)

hcorson-dosch (approved):
Thanks for answering my many questions, Andy. This all looks great to  me.

Issue #10, opened 2022-02-15
Process data to create LSTM training sequences


AndyMcAliley:
Turn drivers, attributes, and temperature observations into fixed-length sequences of LSTM inputs and outputs. Inputs are drivers and attributes (potentially including lake clarity). Outputs are temperatures at preordained depths (and potentially an ice flag indicator).

PR #11, opened 2022-02-15
2_process: Create sequenced training data for LSTM

### LINKED ISSUE #9: Include elevations in lake metadata
We want to use elevation as a lake attribute. It's not included in lake_metadata.csv, so we'll need to add it somehow.

AndyMcAliley:
`2_process` takes raw downloaded observations, drivers and attributes and formats it for model training. Closes #9 and #10.

To run the full pipeline: `snakemake --snakefile Snakefile -c4 -p --rerun-incomplete`. You may choose the number of cores to use in the flag `-c4`.

There are several steps to get the sequences that comprise training data. Each step is associated with a Snakemake rule. Below, I've broken apart each step into the Snakemake rule name, notes on the Python part of the implementation, and notes on the Snakemake part of the implementation.

1. Unzip zipped files to 2_process/tmp folder.
    - `rule: unzip_mntoha`
    - Implementation: 
        - `unzip_all` calls `unzip_file`.
    - Snakemake: 
        - The `get_mntoha_input_files` function provides the list of zip files to unzip.
        - The output is a log file listing all unzipped files. This output file acts like a dummy file. Later rules can include this output log as an input if they depend directly on the unzipped files.
        - This step is not parallelized. It could be, by somehow associating every zip file with an output file, then calling a summary function to write a combined log file. Creating a separate output file for each zip file seems messy, though. Unzipping doesn't take too long and only needs to be done once, so I didn't think it was worth the extra complexity and effort.
2. Interpolate temperature observations from arbitrary depths to a fixed set of discrete depths.
    - `rule: interpolate_mntoha_obs_depths`
    - Implementation:
        - `make_obs_interpolated` uses nearest neighbor interpolation to assign each observation to the nearest discrete depth value.
        - The output file is identical to the input file, but with an added column: "interpolated_depth".
    - Snakemake:
        - The observation file is zipped upon downloading, so this rule includes the `unzip_mntoha` log file as an input to make sure it gets unzipped first.
3. Augment the lake metadata to include all lake attributes to be used as model input features (right now, just add elevation)
    - `rule: augment_mntoha_lake_metadata`
    - Implementation:
        - The output file is identical to the input file, but with an added column: "elevation".
        - We may want to augment the metadata with more lake attributes at this step later. That's why the names of the rule and the function aren't specific to adding elevation.
        - This rule takes a while due to a request to the USGS Elevation Point Query Service for all the elevations at each lake's lat/lon centroid. 
    - Snakemake
        - Unlike `interpolate_mntoha_obs_depths`, `unzip_mntoha`'s log file isn't included, only `fetch_all`'s output `downloaded_files.txt`. That's becaues `lake_metadata.csv` isn't zipped on ScienceBase, so this rule doesn't rely on unzipping.
        - The full path to the metadata csv is `1_fetch/out/metadata_mntoha/lake_metadata.csv`. So, it matches the `1_fetch/out/{file_category}_mntoha/{file}` pattern of `fetch_mtoha_data_file`.
4. Create fixed-length time series sequences of inputs/outputs for one MNTOHA lake. Each sequence will constitute one training example.
    - `rule: mntoha_lake_sequences`
    - Implementation:
        - I tried to write this code so that it will be reusable when we switch to another data source beyond MNTOHA, but I didn't spend much time anticipating how that will work.
        - Each lake's sequences get written to a lake-specific .npy file. Those .npy files can be combined later, as needed.
        - This step is the most computationally costly so far.
        - This step was tricky to implement. There are two particularly tricky aspects to it.
            * Take sparse observations specific to a few depths and a few days, and create a big full array of all depths and days where missing observations are NaN. `all_dates_depths` does the hard work here, using `pandas`.
            * Divide the big full array into equal length sequences of length `sequence_length`. The maybe-too-clever-for-its-own-good bit uses `np.lib.stride_tricks` to do this without having to copy the big full array.
    - Snakemake:
        - Drivers, including clarity and ice flags, are zipped upon download. Therefore, this rule includes `unzip_mntoha`'s log file as an input.
        - I used `script` instead of `run` here, but the rule could be implemented either way. We could choose to standardize and call either `script` or `run` every time, if that would help to simplify things.
5. Execute step 4 for all lakes in parallel.
    - `rule: process_mntoha`
    - Implementation: The Python code is straightforward.
    - Snakemake:
        - The input is every lake sequence file that `mntoha_lake_sequences` should create. That way, when the rule `process_mntoha` is called, it will trigger each lake sequence file to be built.
            * The function `mntoha_lake_sequence_files` reads `lake_metadata.csv` to return that list of every sequence file that should be created.
            * The problem is that we need `lake_metadata.csv` to create that list of lakes. `lake_metadata.csv` is itself downloaded during the `fetch_all` rule, so it's not necessarily available when snakemake is first called. Therefore, Snakemake can't create its full dependency graph up front. To deal with this, I turned the rule `fetch_all` into a [checkpoint](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html?highlight=checkpoint#data-dependent-conditional-execution). See also [here](https://edwards.flinders.edu.au/how-to-use-snakemake-checkpoints/) and [here](http://ivory.idyll.org/blog/2021-snakemake-checkpoints.html) for small introductions to checkpoints.
            * The function `mntoha_lake_sequence_files` calls upon the output of the `fetch_all` checkpoint. This accomplishes two things.
                1. It makes the `process_mntoha` rule dependent upon `fetch_all`.
                2. It lets Snakemake know not to keep running the `mntoha_lake_sequence_files` function until `fetch_all` is complete. Since `mntoha_lake_sequence_files` specifies jobs that `process_mntoha` depends on, Snakemake knows to hold off on adding those jobs to the dependency graph until after `fetch_all` has run.
        - Like `unzip_mntoha`, the output is a summary text file that acts like a dummy file to ensure that all lakes have been processed through `mntoha_lake_sequences`.
        - The function `save_sequences_summary` provides that summary text file.


jsadler2:
@AndyMcAliley - do you think you could upload an image of the DAG? I'm hoping that won't take you long and would make it easier for me to quickly see what your workflow is doing. You can render just the bit in phase `2_` to an `svg` with:
```
snakemake 2_process/out/mntoha_sequences/mntoha_sequences_summary.csv -s 2_process.smk  --dag | dot -Tsvg > dag.svg
```
Then you'd have to convert that to a .png or .jpg with either an online tool (https://svgtopng.com) or something like Inkscape.

The one hitch might come with having the right libraries (like dot and graphviz) so if it's turning into a rabbit hole, don't worry about it.

notebooks/Explore_MNTOHA_temperature_observations.ipynb
```diff
@@ -0,0 +1,344 @@
+{
```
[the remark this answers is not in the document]
jsadler2: If you are familiar with rmarkdown, but jupytext provides roughly the same functionality as rmarkdown (edit in plain text and then it renders it to a notebook).

2_process.smk
```diff
@@ -0,0 +1,106 @@
...
+        config["process_mntoha"]["metadata_augmented_file"],
+        config["process_mntoha"]["obs_interpolated_file"],
+        config["process_mntoha"]["unzip_log_file"]
+    output:
+        "2_process/out/mntoha_sequences/sequences_{site_id}.npy"
+    params:
+        # site_id = lambda wildcards: wildcards.site_id,
+        config = config["process_mntoha"]
+    script:
+        "2_process/src/lake_sequences_mntoha.py"
```
jsadler2, line 72: You mentioned this in your comments. If it were me, I'd either try to stick with all `script` or all `run`. The nice thing about `script` is that snakemake knows when a file has been changed. The bad thing is that you have to write a different file for each step.

jsadler2 (commented):
Hi @AndyMcAliley - I have made some suggestions for you guys to consider that could make the dependency graph a bit cleaner. Some of these are just a different way of thinking about and structuring the workflow. I'm happy to talk through the suggestions or answer any questions if you have them.

2_process.smk
```diff
@@ -0,0 +1,106 @@
...
+    input:
+        config["process_mntoha"]["unzip_log_file"],
+    output:
+        config["process_mntoha"]["obs_interpolated_file"]
+    run:
+        make_obs_interpolated(config["process_mntoha"]["obs_file"], output[0], config["process_mntoha"]["depths"])
+
+
+# Unzip all MNTOHA data downloaded from ScienceBase
+rule unzip_mntoha:
```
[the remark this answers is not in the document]
AndyMcAliley: For posterity: Jeff and I talked through this. Phase `2_process` depends on all files from phase `1_fetch`: lake_metadata.csv, temperature_observations.csv, and all the drivers, clarity, and ice_flags files. I did move the unzip step to `1_fetch`. However, unzipping is tricky because the number of unzipped files varies depending on what's being unzipped. So, I used Snakemake checkpoints for dynamic branching.

2_process.smk
```diff
@@ -0,0 +1,106 @@
...
+    run:
+        make_lake_metadata_augmented(config["process_mntoha"]["metadata_file"], output[0])
+
+
+# Add column of observation depths interpolated to nearest modeling mesh node
+rule interpolate_mntoha_obs_depths:
+    input:
+        config["process_mntoha"]["unzip_log_file"],
+    output:
+        config["process_mntoha"]["obs_interpolated_file"]
```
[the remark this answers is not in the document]
AndyMcAliley: Agreed! I moved file names out of the config files and into the Snakemake files.

2_process.smk
```diff
@@ -0,0 +1,106 @@
...
+
+
+# Add elevation to MNTOHA lake metadata
+rule augment_mntoha_lake_metadata:
+    input:
+        "1_fetch/out/downloaded_files.txt"
+    output:
+        config["process_mntoha"]["metadata_augmented_file"]
+    run:
+        make_lake_metadata_augmented(config["process_mntoha"]["metadata_file"], output[0])
```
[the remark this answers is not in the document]
AndyMcAliley: Yes! Before, that change threw an error because of the way the checkpoints in `1_fetch.smk` were set up. I changed those checkpoints, and now this more straightforward way of relying on lake_metadata.csv as input works.

2_process.smk
```diff
@@ -0,0 +1,106 @@
...
+
+    :param wildcards: Snakemake wildcards (unused).
+    :returns: List of lake training/testing sequence files.
+
+    """
+    # make this function dependent on fetch_all
+    # needed because lake_metadata.csv is used to determine lake_sequence_files
+    ck_output = checkpoints.fetch_all.get(**wildcards).output[0]
+    out_dir = os.path.join(config["process_mntoha"]["out_dir"], 'mntoha_sequences')
+    lake_metadata_file = config["process_mntoha"]["metadata_file"]
```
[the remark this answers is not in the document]
AndyMcAliley: Now that I changed the checkpoints in `1_fetch`, this comment is outdated. I did follow the spirit of it, though, and made `unzip_archive` a checkpoint. That lets Snakemake only download and unzip what it needs.

2_process/process_config.yaml
```diff
@@ -0,0 +1,17 @@
+process_mntoha:
```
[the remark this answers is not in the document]
AndyMcAliley: Done! Now that the config doesn't have any filenames or directories, we don't have a reason for this top level.

1_fetch.smk
```diff
@@ -12,23 +12,29 @@ def get_mntoha_input_files(wildcards):
...
+checkpoint fetch_all:
     input:
         "1_fetch/in/pull_date.txt",
         get_mntoha_input_files
+    output:
+        "1_fetch/out/downloaded_files.txt"
+    run:
+        with open(output[0], 'w') as f:
+            for input_file in input[1]:
+                f.write(input_file + '\n')
```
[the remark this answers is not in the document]
AndyMcAliley: `fetch_all` is gone. It took some thinking to make this work because there's dynamic branching at play. In the old way I was doing it, fetch_all made sure EVERYTHING I knew I'd was was downloaded. It was a checkpoint because I needed to use `lake_metadata.csv` to determine outputs of downstream rules (dynamic branching). In the new way, I have a new checkpoint, `fetch_mntoha_metadata`, which downloads `lake_metadata.csv` only. Later rules that depend on `lake_metadata.csv` for dynamic branching call this checkpoint in their input functions.

The other big change I needed to make had to do with how to unzip files. Lake sequences need unzipped driver, clarity, and ice flags files. The unzipped files should depend on the unzipping rule in the DAG. Unzipping can be a one-to-many operation where the number of outputs varies - dynamic branching! So, I made another checkpoint, `unzip_archive`. This does the work of unzipping, and its output is the directory that files are unzipped to. Then, any rule that relies on unzipped files calls the `unzip_archive` checkpoint in its input function.

notebooks/Explore_MNTOHA_temperature_observations.ipynb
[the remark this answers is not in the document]
AndyMcAliley: Thanks! This looks really neat. It looks like others would need Jupytext installed to easily read and run the notebooks. I'll check with our team to see what they think.

2_process.smk
AndyMcAliley: Changed to `script`! I like that Snakemake can watch for changes to the script.

AndyMcAliley:
New DAG, obtained by running `snakemake --filegraph | dot -Tsvg > filegraph.svg`

filegraph

notebooks/Explore_MNTOHA_temperature_observations.ipynb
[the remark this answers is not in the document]
hcorson-dosch: That seems like a neat tool, particularly if we can make use of the [paired notebook](https://github.com/mwouts/jupytext/blob/main/README.md?plain=1#L87-L132) approach, so that the text version is update whenever changes are made in the notebook (or vice versa). Whenever you've got time for a little side project, it could be worth doing a little test case where you make a notebook Andy, commit the text version, and I test how well it works to render it, make edits, and commit those edits on my end.

1_fetch.smk
```diff
@@ -1,43 +1,59 @@
...
+# Download MNTOHA meteorological drivers, clarity, and ice flag zip files
+rule fetch_mntoha_dynamic:
     input:
         "1_fetch/in/pull_date.txt"
     output:
-        "1_fetch/out/{file_category}_mntoha/{file}"
+        "1_fetch/tmp/dynamic_mntoha/{file}"
     params:
-        sb_id = lambda wildcards: config["mntoha"][wildcards.file_category]["sb_id"]
+        sb_id = "5e5d0b96e4b01d50924f2b34"
```
hcorson-dosch, line 37: okay, so we're now just automatically downloading everything from this `sb_id`, right?

1_fetch/src/unzip_file.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    # check that file is inside source_dir
+    if os.path.samefile(source_dir, os.path.commonpath([filename, source_dir])):
+        relfile = os.path.relpath(filename, source_dir)
+        destination_file = os.path.join(destination_dir, os.path.splitext(relfile)[0])
+        with zipfile.ZipFile(filename, 'r') as zf:
+            unzipped_files = zf.namelist()
+            zf.extractall(destination_file)
+    else:
+        raise FileNotFoundError(f'File {f} not in directory {source_dir}')
```
hcorson-dosch, line 24: should `{f}` here be `{filename}`?

1_fetch/src/unzip_file.py
```diff
@@ -0,0 +1,33 @@
...
+
+    :param filename: File to unzip, with full paths
+    :param source_dir: Directory to pull zipped file from.
+        This part of the path is removed from the beginning of filename to determine where to save unzipped files.
+    :param destination_dir: Directory to save unzipped files to.
+        In each file path, source_dir is replaced with destination_dir to determine where to save unzipped files.
+    :returns: list of paths to unzipped files
+
+    """
+    # check that file is inside source_dir
```
hcorson-dosch, line 16: I think this just checks that the path to `{filename}` is the same as `{source_dir}`, it doesn't check if `{filename}` actually exists in that `{source_dir}`. I tested this with `{filename}` equal to `'1_fetch/in/test.txt'` and `{source_dir}` set to `'1_fetch/in'`, and it passed the if statement, even though  `'1_fetch/in/test.txt'` doesn't exist
```python
filename = '1_fetch/in/test.txt'
source_dir = '1_fetch/in'

if os.path.samefile(source_dir, os.path.commonpath([filename, source_dir])):
    print(f'{filename} exists')
else:
    raise FileNotFoundError(f'File {f} not in directory {source_dir}')


> 1_fetch/in/test.txt exists
```

1_fetch/src/unzip_file.py
```diff
@@ -0,0 +1,33 @@
...
+    if os.path.samefile(source_dir, os.path.commonpath([filename, source_dir])):
+        relfile = os.path.relpath(filename, source_dir)
+        destination_file = os.path.join(destination_dir, os.path.splitext(relfile)[0])
```
hcorson-dosch, line 19: I'd suggest renaming this to `destination_subdir` rather than calling it a file, since it's the directory into which the files will be unzipped, right?

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+
+def mntoha_lake_sequence_files(wildcards):
+    """
+    List all MNTOHA lake sequence .npy files for training and testing.
+
+    :param wildcards: Snakemake wildcards (empty).
+    :returns: List of lake training/testing sequence files.
+
+    """
+    # make this function dependent on fetch_all
```
hcorson-dosch, line 16: `fetch_all` is no longer a rule, right?

1_fetch.smk
AndyMcAliley: Not everything - only the file requested.

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+    # needed because lake_metadata.csv is used to determine lake_sequence_files
+    lake_metadata_file = checkpoints.fetch_mntoha_metadata.get(**wildcards).output[0]
```
hcorson-dosch, line 18: Just to make sure I'm interpreting this correctly --- here, we're checking if the `fetch_mntoha_metadata` checkpoint has been executed for _everything_ that falls under that checkpoint rule, because of that `**wildcards` notation, right, even though `wildcards` is empty in this case, as you note in the function documentation. And if `wildcards` weren't empty, and there were one named `category`, we could access the output files associated with a category named 'minnesota' using `checkpoints.fetch_mntoha_metadata.get(category="minnesota").output`, right? For reference, I'm reading [these docs](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html) to try to piece it all together.

hcorson-dosch: Or are we just checking that the checkpoint has been executed for that first (and in this case, only) output file `1_fetch/out/lake_metadata.csv`? But, in theory if there were multiple wildcards, it would check that the first output file had been generated for each of them?

Sorry for the snakemake Qs - I know that's not the point of my review -- just trying to get oriented to the pipeline.

1_fetch.smk
hcorson-dosch: Oh okay -- where is the file requested now that there is no `configfile`?

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+import pandas as pd
+
+configfile: "2_process/process_config.yaml"
+
+
+def mntoha_lake_sequence_files(wildcards):
+    """
+    List all MNTOHA lake sequence .npy files for training and testing.
+
+    :param wildcards: Snakemake wildcards (empty).
```
hcorson-dosch, line 12: Thanks for noting that `wildcards` is empty - that's helpful

2_process.smk
AndyMcAliley: Right - that comment is out of date. It should read: 
```
# make this function dependent on fetch_mntoha_metadata
```

2_process/src/make_lake_metadata_augmented.py
```diff
@@ -0,0 +1,52 @@
...
+    params = {
+        'output': 'json',
+        'x': lon,
+        'y': lat,
+        'units': 'Meters'
+    }
+
+    # format query string and return query value
+    result = requests.get((url + urllib.parse.urlencode(params)))
+    elevation = result.json()['USGS_Elevation_Point_Query_Service']['Elevation_Query']['Elevation']
```
hcorson-dosch, line 29: This is so cool!

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+
+    Depend on unzip_archive checkpoint to ensure that
+    temperature_observations.csv gets unzipped.
+
+    :param wildcards: Snakemake wildcards (empty).
+    :returns: Path of MNTOHA temperature observations csv
+
+    '''
+    # Trigger checkpoint to unzip temperature_observations.csv
+    obs_file_directory = checkpoints.unzip_archive.get(
```
hcorson-dosch, line 137: This is pretty nifty!

2_process.smk
AndyMcAliley: The `**wildcards` notation means that whatever wildcards were passed as kwargs into `mntoha_lake_sequence_files` are getting passed on through as kwargs to the checkpoint `fetch_mntoha_metadata`. You're right that `wildcards` is empty here. See `dynamic_filenames` for a case where `wildcards` isn't empty. Yes, you could pass specific values for wildcards to a checkpoint as you wrote. Again, that happens in `dynamic_filenames`.

AndyMcAliley: In `dynamic_filenames`, even though `wildcards` isn't empty, it's unused by the checkpoint `fetch_mntoha_metadata` because there's no wildcards defined in that checkpoint.

2_process.smk
hcorson-dosch: Just so I'm clear -- the way this system works is that `1_fetch/tmp/obs_mntoha/temperature_observations.zip` doesn't get unzipped until the checkpoint is triggered here?

2_process/src/make_lake_metadata_augmented.py
AndyMcAliley: Yes, it's a cool service! I will note, though, that querying 881 elevations for 881 lakes is pretty slow. Jordan's working on adding elevation to the metadata in lake-temperature-model-prep so this query won't be needed for the full set of lakes.

2_process.smk
hcorson-dosch: Ok, cool, that make sense. Thanks for the explanation!

2_process.smk
AndyMcAliley: Yes! The pattern here is a simpler version of the pattern involving `dynamic_filenames` and `mntoha_lake_sequences`. It allows dynamic branching. This is the usage pattern for checkpoints that took me some time to wrap my head around!

1. A rule's input is a function.
2. The function triggers a checkpoint. Snakemake knows now that the rule depends on execution of the checkpoint.
3. The checkpoint creates some files.
4. After the checkpoint is run, Snakemake recreates the pipeline graph and reruns the function.
5. The output of the function are the files that are inputs to the rule. Those files will probably include at least some of the files created by the checkpoint.
6. Snakemake's newly recreated graph can include newly created files.
7. Now that its input files have been created, the rule can run.

1_fetch/src/unzip_file.py
AndyMcAliley: Yes, good point! All it's checking is if the path to `filename` includes `source_dir`. 

I think that if the file itself doesn't exist, then `zipfile` will raise its own error. So, I don't think there's value in actually checking for the file itself. Instead, what if I update the comment and the error message to reflect the actual error checking that's happening here?

hcorson-dosch: that sounds good!

1_fetch.smk
AndyMcAliley: The file is inferred from the output. So, if I run the command
```
snakemake -c1 1_fetch/tmp/dynamic_mntoha/ice_flags_01_N48.00-49.50_W89.50-97.25.zip
```
then only that single zip file is downloaded from ScienceBase. 

1. Snakemake matches the filename to the pattern in `output` of the rule `fetch_mntoha_dynamic`.
2. Snakemake runs the script `sb_fetch.py`
3. That script calls `main` with two arguments: (1) the ScienceBase ID taken from `params` of the rule, and (2) the zip file name taken from `output` of the rule.

hcorson-dosch: Ahh I see now. I was thinking that, as in the last PR, all files would be downloaded in order to 'complete' the rules outlined in `1_fetch.smk`, but now I see that for these files, as with the temp obs, snakemake is not actually downloading the files until they are needed by rules in `2_process.smk`. So essentially `1_fetch.smk` provides the 'recipe' for how to get the files, but doesn't actually download and unzip files until a rule in `2_process.smk` has input that then triggers one of the `1_fetch.smk` 'recipes' -- does that sound right?

AndyMcAliley: In the full pipeline, though, `unzip_archive` ends up calling `fetch_mntoha_dynamic` and providing the file name. The DAG above shows that relationship. `mntoha_lake_sequences` depends on `unzip_archive`, and `mntoha_lake_sequences`'s input function `dynamic_filenames` is truly where the file name is built.

AndyMcAliley: Yes, what you said is right!

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+    unzip_archive_ice_flags = checkpoints.unzip_archive.get(
+        file_category=file_category,
+        directory_name=ice_flags_directory
+    ).output[0]
+
+    # dynamic filenames
+    dynamic_dir = '1_fetch/out'
+    drivers_file = f'{dynamic_dir}/dynamic_mntoha/inputs_{lake.group_id}/{lake.meteo_filename}'
+    clarity_file = f'{dynamic_dir}/dynamic_mntoha/clarity_{lake.group_id}/gam_{lake.site_id}_clarity.csv'
+    ice_flags_file = f'{dynamic_dir}/dynamic_mntoha/ice_flags_{lake.group_id}/pb0_{lake.site_id}_ice_flags.csv'
```
hcorson-dosch, line 87: Since you've already defined parts of these paths as variables, I wonder if would make sense to edit these 3 lines to be:
```python
drivers_file = f'{dynamic_dir}/{file_category}/{drivers_directory}/{lake.meteo_filename}'
clarity_file = f'{dynamic_dir}/{file_category}/{clarity_directory}/gam_{lake.site_id}_clarity.csv'
ice_flags_file = f'{dynamic_dir}/{file_category}/{ice_flags_directory}/pb0_{lake.site_id}_ice_flags.csv'
```

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+    run:
+        save_sequences_summary(input, output[0])
+
+
+# Create .npy of input/output sequences for one lake to use for training and testing
+rule mntoha_lake_sequences:
+    input:
+        "2_process/tmp/mntoha/lake_metadata_augmented.csv",
+        "2_process/tmp/mntoha/temperature_observations_interpolated.csv",
+        dynamic_filenames
```
hcorson-dosch, line 106: For the sake of readability and avoiding hard-coding within functions I find myself wishing there were some way to specify parameters for these functions that make use of checkpoints here in the `smk` file -- e.g., that file_category = 'dynamic_mntoha'.

AndyMcAliley: From the docs [here](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#input-functions-and-unpack), it looks like this can be achieved. Rather than passing the function `dynamic_mntoha`, we could try calling the function and using Python's star operator to unpack the list of file names, like this:
```
input:
    "2_process/tmp/mntoha/lake_metadata_augmented.csv",
    "2_process/tmp/mntoha/temperature_observations_interpolated.csv",
    *dynamic_filenames(file_category='dynamic_mntoha')
```

2_process.smk
AndyMcAliley: Yes, it does! In fact, the outputs from the calls to the `unzip_archive` checkpoint can be used, requiring fewer variables. Change made!

2_process.smk
AndyMcAliley: Indeed, wildcards weren't passed. Instead, using a `lambda` function allowed a keyword to be specified. 
```
    input:
        "2_process/tmp/mntoha/lake_metadata_augmented.csv",
        "2_process/tmp/mntoha/temperature_observations_interpolated.csv",
        lambda wildcards: dynamic_filenames(wildcards.site_id, file_category='dynamic_mntoha')
```
Also, then, the function `dynamic_filenames` doesn't have to bother with the `wildcards` variable. The arguments are now simply the arguments that `dynamic_filenames` needs.

hcorson-dosch: Ooh nice pipelines upgrade, Andy!

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,202 @@
...
+    Process raw data for one MNTOHA lake
+    Save those data to a numpy file
+
+    :param site_id: site_id of lake to process
+    :param out_file: Filename of npy file to save
+    :param metadata_augmented_file: lake metadata file, augmented with elevation
+    :param obs_interpolated_file: temperature observations file, interpolated
+        to LSTM depths
+    :param drivers_file: Meteorological drivers csv for this lake
+    :param clarity_file: Time-varying lake clarity csv
```
hcorson-dosch, line 176: The clarity file is also 'for this lake', correct?

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,202 @@
...
+    :param ice_flags_file: Time-varying ice flags csv for this lake
+    :param config: Snakemake config for process_mntoha
```
hcorson-dosch, line 178: The config is for `2_process` - maybe you had it named `process_mntoha` previously?

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+        mntoha_lake_sequence_files
+    output:
+        "2_process/out/mntoha_sequences/mntoha_sequences_summary.csv"
+    run:
+        save_sequences_summary(input, output[0])
+
+
+# Create .npy of input/output sequences for one lake to use for training and testing
+rule mntoha_lake_sequences:
+    input:
```
hcorson-dosch, line 103: I'm still in the midst of reviewing `assemble_lake_data()`, but I see that while the drivers, clarity, and ice flag files returned by `dynamic_filenames()` are site-specific (since `snakemake` is mapping over the wildcard `site_id` to retrieve those files, the other two inputs, `"2_process/tmp/mntoha/lake_metadata_augmented.csv"` and `"2_process/tmp/mntoha/temperature_observations_interpolated.csv"` are not site-specific. Instead, those two dataframes get filtered to the appropriate site within `assemble_lake_data()`.

Two factors to consider re: the validity of my comment here: 1) I'm not fully up to speed on `snakemake` dependency tracking or what options would be on the table, and 2) this may be less relevant for a global model pipeline than it is for a site-specific model pipeline.

My thoughts: in a `targets` modeling workflow, we'd typically avoid passing in the _complete_ set of site metadata and _complete_ set of observations, because a change in the metadata or # of observations for _any_ site would then trigger a re-build of the sequences file file for _every_ site. Ditto if you added a new site. Instead I think we'd create an intermediary target for the metadata and another for the observations that map over the site_ids to generate site-specific objects/files that would then be passed into `assemble_lake_data()` alongside the other 3 files. I'm not certain what the equivalent approach would be in `snakemake` - maybe you would make a function or functions that would make use of the `site_id` wildcard to filter the observations/metadata to that site, save to a csv for that site, and return that site-specific csv? And then you could list that function(s) under inputs, here?

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Correct! I changed the text.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Yes, you're right! The `2_process` config used to have a subsection called `process_mntoha`, but I flattened that out over the course of this PR. Fixed.

2_process.smk
AndyMcAliley: Yes, that would speed up the pipeline whenever metadata or observations changed. I think you're right that this would make a bigger difference for training many site-specific models - adding new sites wouldn't require that the old sites get re-run.

Snakemake may or may not support that kind of workflow, but so far I haven't learned how to implement that in Snakemake. I think (big caveat: I'm still learning) that Snakemake looks at a file's modification date, and not its contents. So, say that our pipeline separates `lake_metadata.csv` into many site-specific files. Then, say we add a new lake to `lake_metadata.csv`. The function that breaks the csv apart into site-specific files would still overwrite all those files, even though it actually only needed to make one new file. Then, Snakemake would re-execute every downstream task.

So, I think it would take a lot of manual work outside of Snakemake to set up a pipeline in which new files that need to be remade are remade, and files that don't need to be remade are not.

Something like: 
1. Read in every existing file
2. Check to see if the new content that would be written to the file match the old content
3. Only write that new content to the file if they don't match.

hcorson-dosch: Ah okay, that's helpful context, and good to be aware of the constraints with snakemake. Definitely does not seem worth exploring at this point. Thanks!

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+        "2_process/tmp/mntoha/lake_metadata_augmented.csv",
+        "2_process/tmp/mntoha/temperature_observations_interpolated.csv",
+        lambda wildcards: dynamic_filenames(wildcards.site_id, file_category='dynamic_mntoha')
+    output:
+        "2_process/out/mntoha_sequences/sequences_{site_id}.npy"
+    params:
+        config = config
```
hcorson-dosch, line 110: Wait I'm confused -- I see that the configuration file is referenced here and on line 152 through the variable `config`, but where the variable for `"2_process/process_config.yaml"` is defined on line 5, the variable is named as `configfile`. Does `snakemake` know that config = configfile, or did you maybe switch the variable name you were using at one point?

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,202 @@
...
+
+    pad_before_days_td = np.timedelta64(pad_before_days, 'D')
+    pad_after_days_td = np.timedelta64(pad_after_days, 'D')
+    min_days_td = np.timedelta64(min_days, 'D')
+    one_day = np.timedelta64(1, 'D')
+    # buffer start_date by pad_before_days days
+    start_date = lake_obs_interpolated.date.min() - pad_before_days_td
+    # buffer end_date by pad_after_days days
+    end_date = lake_obs_interpolated.date.max() + pad_after_days_td
+    # check if enough days are included, pad if not
```
hcorson-dosch, line 28: maybe just make clear in your comment here that you're padding to the left, i.e., adding days to the start of the sequence to ensure it is long enough, rather than adding days to the end of the sequence

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,202 @@
...
+    if end_date - start_date + one_day < min_days_td:
+        start_date = end_date - min_days_td + one_day
+    date_range = pd.date_range(start=start_date, end=end_date)
+
+    # create MultiIndex of all combinations of dates and depths
+    midx = pd.MultiIndex.from_product([date_range, depths], names=['date', depth_col])
+    # Average duplicate measurements (same day, same nearest depth)
+    obs_midx = lake_obs_interpolated.groupby(['date', depth_col]).temp.mean().to_frame()
+    obs_all = obs_midx.reindex(midx, fill_value=np.NaN)
```
hcorson-dosch, line 37: I'd just add a comment here to note that you're expanding to the full set of date ranges and depths set by the multi-index and filling w/ NAs for date-depth instances where there is no observation

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,202 @@
...
+
+    # reshape to have one column per depth
+    obs_all = obs_all.unstack()
+    # flatten column levels
+    obs_all.columns = [' '.join([str(s) for s in col]).strip() for col in obs_all.columns.values]
```
hcorson-dosch, line 42: Phew took me a bit to unpack what was happening here - had to go mock it up in a notebook. I'd forgotten that `unstack()` generates those weird column levels. I think this could be written more simply by iterating through both elements of the column names:
```python
obs_all.columns = [f'{i} {j}' for i,j in obs_all.columns.values]
```
And I'd suggest adding a note in your comment that `unstack()` generates two column levels: `temp` and the interpolated depths, and it's those values that are being flattened here to 'temp 0.0', etc. Also - this is nitpicking, esp. b/c the column names get dropped later when you convert the dataframe to a numpy array, but typically we try to avoid spaces in column names and use underscores instead.

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,202 @@
...
+
+        # Fill NaNs for missing obs
+        obs_all = all_dates_depths(lake_obs_interpolated,
+                                   depths,
+                                   pad_before_days=spinup_time,
+                                   min_days=sequence_length)
+        # Now obs_all has dimensions of len(date_range) by len(depths)
+        # values are NaN wherever there are no observations
+
+        # Join drivers
```
hcorson-dosch, line 101: For each of these joins I'd suggest just adding a note that you're joining by date, since you're defaulting to the `join()` option to join on the index of each dataframe.

2_process.smk
AndyMcAliley: That's the syntax that Snakemake uses for [configuration](https://snakemake.readthedocs.io/en/latest/snakefiles/configuration.html) 🤷🏻.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Thanks for that snippet - so much nicer! I went through `all_dates_depths` and rewrote the comments to improve readability.

2_process.smk
hcorson-dosch: huh - bizarre!

2_process/src/lake_sequences_mntoha.py
hcorson-dosch: Thanks Andy - these new comments are super helpful

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+    # If not, pad to the left (shift the start date earlier)
+    if end_date - start_date + one_day < min_days_td:
+        start_date = end_date - min_days_td + one_day
+    date_range = pd.date_range(start=start_date, end=end_date)
+
+    # Now take sparse observations specific to a few depths and a few days, 
+    # and create a full DataFrame of all depths and days, with NaN values for missing observations.
+
+    # Take only the columns we need
+    obs = lake_obs_interpolated[[date_col, depth_col, temp_col]]
```
hcorson-dosch, line 47: Oh I like this edit - it's much easier to follow and also makes your intentions plain

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+    # Average duplicate measurements (same day, same nearest depth)
+    # The resulting DataFrame has a MultiIndex of (date, depth)
+    obs_midx = obs.groupby([date_col, depth_col]).mean()
+    # Create a full MultiIndex of all possible combinations of dates and depths
+    midx = pd.MultiIndex.from_product([date_range, depths], names=[date_col, depth_col])
+    # Expand to the full set of date ranges and depths set by the MultiIndex 
+    # Values are NaN for date-depth combinations that have no observations
+    obs_full = obs_midx.reindex(midx, fill_value=np.NaN)
```
hcorson-dosch, line 55: I also like this new more descriptive variable name!

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+                                    pad_before_days=spinup_time,
+                                    min_days=sequence_length)
+        # Now obs_full has dimensions of len(date_range) by len(depths)
+        # Values are NaN wherever/whenever there are no observations
+
+        # There are some NaN values in the clarity and ice flag data.
+        # There are multiple ways to deal with NaN values in inputs:
+        # interpolate to fill them, skip over them, or remove sequences
+        # containing NaNs entirely. At this point, allow NaN values into the
+        # sites. Later code will handle NaN values.
```
hcorson-dosch, line 126: This comment block is a nice addition. I really like comments like this that orient the user to choices being made.

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        # 1. Sequences are sequence_length days long
+        # 2. Each next sequence starts sequence_offset days later than the previous
+        # 3. Discard sequences that lack observations
+        # 4. We know going in that the first and last sequences have observations
+
+        # Convert dataframe to numpy array
+        obs_full_array = obs_full.to_numpy(dtype=np.float32)
+        # Create a strided view into obs_full_array
+        # This creates a numpy object with sequences of length sequence_length
+        # and offsets of sequence_offset, without making an in-memory copy of the array
```
hcorson-dosch, line 160: what a cool function!

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        all_sequences = np.lib.stride_tricks.sliding_window_view(
+            obs_full_array, sequence_length, axis=0
+        )[::sequence_offset]
+        # The shape of all_sequences is (# sequences, # features + depths, sequence_length)
```
hcorson-dosch, line 164: I might just spell out here that `features` = drivers + clarity + ice flags (right?), since I think that is the first time you're using the using the term `features` in the body of the function

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        # Split into sequences
+
+        # Let's make some rules for sequences:
+        # 1. Sequences are sequence_length days long
+        # 2. Each next sequence starts sequence_offset days later than the previous
+        # 3. Discard sequences that lack observations
+        # 4. We know going in that the first and last sequences have observations
+
+        # Convert dataframe to numpy array
+        obs_full_array = obs_full.to_numpy(dtype=np.float32)
```
hcorson-dosch, line 157: I think it would be helpful to note here, before we create `all_sequences` what the shape of `obs_full_array` is - I think it's (#dates, # features (drivers, clarity, ice_flags) + #depths), right?

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        # Create a strided view into obs_full_array
+        # This creates a numpy object with sequences of length sequence_length
+        # and offsets of sequence_offset, without making an in-memory copy of the array
+        all_sequences = np.lib.stride_tricks.sliding_window_view(
+            obs_full_array, sequence_length, axis=0
+        )[::sequence_offset]
+        # The shape of all_sequences is (# sequences, # features + depths, sequence_length)
+        # The first `len(depths)` elements in the middle dimension are observations
```
hcorson-dosch, line 165: since `depths` come first in that second dimension, I'd change all references to that dimension, e.g., line 164, to make it clear that depth values precede the features in that dimension, rather than the other way around. Also I'd just clarify that you mean _temperature_ observations here

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        lake_sequences = np.empty((0))
+    else:
+        # Read metadata with elevation
+        lake_metadata_augmented = pd.read_csv(metadata_augmented_file)
+        # NOTE: If there are duplicate sites in lake_metadata_augmented,
+        # only the first will be used.
+        # If there are no matches, an IndexError will be raised.
+        lake = lake_metadata_augmented[lake_metadata_augmented.site_id==site_id].iloc[0,:]
+
+        # Create one large DataFrame of temperatures at all days and all discretized depths
```
hcorson-dosch, line 114: I might make clear here that 'all days' is not all dates with observations, but a sequence of dates (`date_range`, as you refer to it in line 119 and in `all_dates_depths()`) generated by padding the range of observed dates to a) provide spin-up time and b) ensure that the date sequence (`date_range`) is at least `sequence_length` long

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        obs_full = all_dates_depths(lake_obs_interpolated,
+                                    depths,
+                                    pad_before_days=spinup_time,
+                                    min_days=sequence_length)
+        # Now obs_full has dimensions of len(date_range) by len(depths)
+        # Values are NaN wherever/whenever there are no observations
```
hcorson-dosch, line 120: These are helpful notes, but I'd move them to before you call `all_dates_depth()`, so that the reader knows that the function is being called specifically to a) expand to `date_range` and b) fill with NaN where there aren't observations.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Good call. I'm using "features" to mean drivers, clarity, ice flags, AND static attributes. I clarified that.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: That's right. I added a comment.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Yes, thanks! I had the ordering of depths and features backwards in my head at one time, and never updated the comments. Observations clarified to mean temperatures, here and elsewhere.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Well put! I'm using your phrasing in the updated comments.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Done! I also added more info to the docstring of `all_dates_depths`.

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,228 @@
...
+        # 4. We know going in that the first and last sequences have observations
+
+        # Convert dataframe to numpy array
+        obs_full_array = obs_full.to_numpy(dtype=np.float32)
+        # Create a strided view into obs_full_array
+        # This creates a numpy object with sequences of length sequence_length
+        # and offsets of sequence_offset, without making an in-memory copy of the array
+        all_sequences = np.lib.stride_tricks.sliding_window_view(
+            obs_full_array, sequence_length, axis=0
+        )[::sequence_offset]
```
hcorson-dosch, line 176: Okay @AndyMcAliley I just want to confirm that I understand what is happening here.

Here's my take:

The first bit `np.lib.stride_tricks.sliding_window_view(obs_full_array, sequence_length, axis=0)` is generating a sliding window 'view' into the `obs_full_array`. `obs_full_array` is shape (`date_range`, #depths + #features). Because you are specifying an integer (`sequence_length`) as the `window_shape` argument to `sliding_window_view()` and `axis=0`, the sliding window is shape `(sequence_length, )` and slides vertically across the rows (dates), generating a window that starts with the first date, then another that starts with the second date, and so on. The shape of the 'view' returned by that function call is (#windows, #depths + #features, `sequence_length`).

The last bit `[::sequence_offset]` pulls elements from the 3D returned 'view' array along the first dimension (#windows) with a step size of `sequence_offset`. So the first window -- with shape (#depths + #features, `sequence_length`) -- is pulled and then every nth window (where n = `sequence_offset`) is pulled until the end of the #windows dimension is reached.

In the end, we have an array with shape (# sequences, #depths + # features, `sequence_length`), where each sequence is offset from the previous sequence by `sequence_offset` days).

If that is a correct interpretation, I wonder if you might change that last bit to read `[::sequence_offset,:,:]`. I know it's not technically  necessary, but might make it clearer that the step indexing is only being applied to the first dimension of the returned array (#windows), and that the other dimensions (#depths + #features and `sequence_length`) are being pulled in full.

2_process/src/lake_sequences_mntoha.py
hcorson-dosch: oh right, I forgot about the static attributes. Thanks for adding that!

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Yes, that's it! Sure, I can add the other dimensions in the slice operation.

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,241 @@
...
+        # The shape of all_sequences is (# sequences, # depths + # features, sequence_length).
+
+        # Only write sequences to file if they contain temperature observations
+        sequences_have_no_obs = np.all(np.isnan(all_sequences[:, :len(depths), :]), axis=(1, 2))
```
hcorson-dosch, line 180: might just note again here that "# The first `len(depths)` elements in the second dimension are temperature observations."

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,241 @@
...
+        # Sequences to save
+        lake_sequences = all_sequences[sequences_have_obs, :, :].transpose((0, 2, 1))
+        # After transpose, shape is (# sequences, sequence_length, # depths + # features)
+
+        # The final temperature can be non-NaN, but the strided view's
+        # sequences probably don't include the final day of the time period due
+        # to truncation. So, add one more sequence that contains the final
+        # sequence_length days, unless all_sequences happened to line up
+        # perfectly such that the last sequence included that final day.
+        if (obs_full_array.shape[0] - sequence_length) % sequence_offset != 0:
```
hcorson-dosch, line 191: Maybe just restate here that the shape of `obs_full_array` is (`date_range`, #depths + #features + #static attributes)

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+
+    :param wildcards: Snakemake wildcards (empty).
+    :returns: List of lake training/testing sequence files.
+
+    """
+    # make this function dependent on fetch_mntoha_metadata
+    # needed because lake_metadata.csv is used to determine lake_sequence_files
+    lake_metadata_file = checkpoints.fetch_mntoha_metadata.get(**wildcards).output[0]
+    lake_metadata = pd.read_csv(lake_metadata_file)
+    lake_sequence_files = [os.path.join('2_process/out/mntoha_sequences', f'sequences_{site_id}.npy') for site_id in lake_metadata.site_id]
```
hcorson-dosch, line 20: Hmm one thing that doesn't seem ideal is that if you changed the filename template on line 108, it wouldn't match what is defined here 🤔

Just thinking out loud here -- I wonder if it would be more robust to do this: 

switch line 94 from ` mntoha_lake_sequence_files` to
```python
lambda wildcards: mntoha_lake_sequence_files(sequence_file_template='2_process/out/mntoha_sequences/sequences_{}.npy')
```

and then switch this line to
``` python
lake_sequence_files = [sequence_file_template.format(site) for site_id in lake_metadata.site_id]
```
and add `sequence_file_template` as an argument to `mntoha_lake_sequence_files()`

That would pull the file template down into the 'rules' section, so would make it a little less hidden, though it still would need to be updated manually if line 108 changed. 🤷‍♀️ What do you think?

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+    Summarize the number of sequences with at least one temperature observation
+    for each lake, and save the result to csv
+
+    :param lake_sequence_files_input: the lake sequence files to summarize
+    :param summary_file: csv file with how many sequences are in each lake
+
+    """
+    sequence_counts = []
+    for sequences_file in lake_sequence_files_input:
+        num_sequences = np.load(sequences_file).shape[0]
```
hcorson-dosch, line 35: Might just note here that the arrays saved in the sequence files have shape (# sequences, `sequence_length`, # depths + # features)

2_process/src/lake_sequences_mntoha.py
```diff
@@ -0,0 +1,241 @@
+import numpy as np
+import pandas as pd
+
+def all_dates_depths(lake_obs_interpolated, 
+                     depths,
+                     temp_col='temp',
+                     depth_col='interpolated_depth',
+                     date_col='date',
```
hcorson-dosch, line 8: Okay parting thought - I've been reflecting on it, and if these 3 column name parameters could change with different data sources, it feels a little brittle to have them hidden as default arguments. I'm not sure if this would violate `snakemake` best practices, but it would feel more in keeping with our standard pipeline practices of documenting all key parameters and decisions in the main 'recipe' file if these columns were specified as `params` to the `rule mntoha_lake_sequences` rule, then passed through as arguments to `assemble_lake_data()` and then as arguments to `all_dates_depth()`.

Making that change here may not be warranted/may not make sense, but as you keep developing the pipeline, I think it would be worth exploring specifying more arguments with `params`, like you ended up doing with the sb_ids in `1_fetch.smk`. I think it would make the `snakemake` structure a bit closer to the structure of our `targets` pipelines.

2_process.smk
```diff
@@ -0,0 +1,155 @@
...
+        directory_name='temperature_observations'
+    ).output[0]
+    return os.path.join(obs_file_directory, "temperature_observations.csv")
+
+
+# Add column of observation depths interpolated to nearest modeling mesh node
+rule interpolate_mntoha_obs_depths:
+    input:
+        # "1_fetch/out/obs_mntoha/temperature_observations/temperature_observations.csv"
+        mntoha_obs_file
```
hcorson-dosch, line 148: Okay I had a thought how you could make the use of the unzip checkpoint a bit more modular - assuming this approach would still make `dynamic_filenames()` and `mntoha_obs_file()` depend on `unzip_archive` in `snakemake`'s eyes.

You could add a function to this `.smk` that just calls the checkpoint, like so:
```python
def return_unzipped_archive(file_category, directory_name):
   # Trigger checkpoint to unzip zipped directory
    unzipped_archive = checkpoints.unzip_archive.get(
        file_category=file_category,
        directory_name=directory_name
    ).output[0]
    return(unzipped_archive)
```

Then you could change line 148 to 
```python
lambda wildcards: get_mntoha_obs_file(wildcards, file_category='obs_mntoha', directory_name='temperature_observations', file_name='temperature_observations.csv')
```

And change `mntoha_obs_file()` to 
```python
def get_mntoha_obs_file(wildcards, file_cateogory, directory_name, file_name):
    obs_file_directory = return_unzipped_archive(file_category, directory_name)
    return os.path.join(obs_file_directory, file_name)
```

And you could update line 106 to:
```ptyhon
lambda wildcards: get_dynamic_filenames(wildcards.site_id, file_category='dynamic_mntoha', drivers_dir_template =''inputs_{}", clarity_dir_template="clarity_{}", ice_dir_template="ice_flags_{}")
```
    
and `dynamic_filenames()` to
```python
def get_dynamic_filenames(site_id, file_category = 'dynamic_mntoha'):
    # make this function depend on fetch_mntoha_metadata
    # needed because lake_metadata.csv is used to determine dynamic files
    lake_metadata_file = checkpoints.fetch_mntoha_metadata.get().output[0]
    lake_metadata = pd.read_csv(lake_metadata_file)
    lake = lake_metadata.loc[lake_metadata['site_id']==site_id].iloc[0]
    # also make this function depend on unzip_archive
    # needed to link unzipped files with unzip_archive rule
    drivers_directory = drivers_dir_template.format(lake.group_id)
    unzip_archive_drivers = return_unzipped_archive(file_category, directory_name=drivers_directory)
    clarity_directory = clarity_dir_template.format(lake.group_id)
    unzip_archive_clarity = return_unzipped_archive(file_category, directory_name=clarity_directory)
    ice_flags_directory = ice_dir_template.format(lake.group_id)
    unzip_archive_ice_flags = return_unzipped_archive(file_category, directory_name=ice_flags_directory)

    # dynamic filenames
    drivers_file = f'{unzip_archive_drivers}/{lake.meteo_filename}'
    clarity_file = f'{unzip_archive_clarity}/gam_{lake.site_id}_clarity.csv'
    ice_flags_file = f'{unzip_archive_ice_flags}/pb0_{lake.site_id}_ice_flags.csv'
    return [drivers_file, clarity_file, ice_flags_file]
```

hcorson-dosch (approved):
Alright! I made it through - I apologize that my review was drawn out over so many days 😅

First off, major kudos to you 👏 for figuring out how to implement this complex workflow in `snakemake` and make use of the checkpoints to allow dynamic branching. You'll have some great `snakemake` insight to share with the DSP group. It was fun for me to wrap my head around a different pipelining package and approach! And your numpy operations in `2_process/src/lake_sequences_mntoha.py` are compact and well commented, and that `sliding_window_view()` function is pretty cool.

You've already made a lot of edits based on my comments, and I think the pipeline and your code are looking really good. I appreciate you detailed comments and your responses to my many questions.

I made 3 parting comments with options for modularizing the code a bit and specifying some additional arguments in the `2_process.smk` file itself - you can see if those ideas make sense to you and seem worth implementing or no.

But all in all this looks really great, and the only remaining required edits based on my review are the few small remaining edits to comments that I noted.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: That sounds good. I like the idea of capturing them in the recipe file.

2_process.smk
AndyMcAliley: Yes, keeping the filename template in the rules and not hardcoded in functions would be a more robust solution. I think that will set us up better for the future, too, when we incorporate data beyond MNTOHA.

2_process.smk
AndyMcAliley: Nice! I'll test this out.

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Done!

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: Done!

2_process/src/lake_sequences_mntoha.py
AndyMcAliley: It's fancy!

AndyMcAliley:
I also reordered the rules and functions in `2_process.smk` so that their ordering follows the order they're executed in the pipeline. Closing now!

PR #21, opened 2022-03-25
Save training and test data

### LINKED ISSUE #10: Process data to create LSTM training sequences

AndyMcAliley:
So far, sequences are saved in separate .npy files, one per lake. This PR takes those separate .npy files and turns them into two new .npz files: one for training data and one for test data. I had planned to do this step in `3_train`, but after [this discussion](https://github.com/USGS-R/lake-temperature-lstm-static/discussions/14) I was persuaded to move it to `2_process` because it aligns better with the goal of `2_process` - process the data before training the model.

Training and test data are split apart randomly by lake. In other words, the sequences from one lake are not divided among the training and test sets - they all either go into the training set or the test set. The idea behind this is to use the test set to evaluate how well the LSTM generalizes to lakes that are not in the training set. 

The new Snakemake rule is `create_training_data`. It calls the script `2_process/src/training_data.py`. Most of the magic happens in the function `get_train_test_data()`.

The script:
- reads sequence files

- discards sequences with `NaN` values in the input features

- splits training and test data 

- standardizes all the features

- saves `2_process/out/mntoha/train.npz` and `2_process/out/mntoha/test.npz`

# Code to run the pipeline and test

To run the pipeline and create training data:
```
snakemake --cores all "2_process/out/mntoha/train.npz"
```

If you want to explore the resulting files, fire up python and run
```
import numpy as np
train_npz = np.load("2_process/out/mntoha/train.npz")
```

Closes #10.


2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+        snakemake.params['test_frac'],
+        snakemake.params['n_depths'],
+        snakemake.params['n_dynamic'],
+        snakemake.params['n_static'],
+        snakemake.params['seed']
+    )
+    # save data to npz along with metadata
+    npz_file_train = snakemake.output['train_file']
+    npz_file_test = snakemake.output['test_file']
+    # **snakemake.params saves all snakemake params to npz, using their names
```
hcorson-dosch, line 153: that's handy

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    containing all sequences from all sequence files.
+
+    :param lake_sequence_files: List of lake sequence .npy files
+    :param n_depths: Number of discrete depths at which LSTM will
+    output temperatures
+    :returns: numpy array of sequences
+
+    """
+    sequences = np.concatenate([np.load(sfn) for sfn in lake_sequence_files])
+    # remove inputs with nans
```
hcorson-dosch, line 67: Maybe make clear here that you're removing inputs with nans for _features_. At first I thought you were again checking for nans within the observation data.

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    NaN values among their inputs, and return a single numpy array
+    containing all sequences from all sequence files.
+
+    :param lake_sequence_files: List of lake sequence .npy files
+    :param n_depths: Number of discrete depths at which LSTM will
+    output temperatures
+    :returns: numpy array of sequences
+
+    """
+    sequences = np.concatenate([np.load(sfn) for sfn in lake_sequence_files])
```
hcorson-dosch, line 66: Also want to make sure that I understand what is happening here. Up until this point, we were computing and storing sequences for each lake separately. the `"2_process/out/{data_source}_sequences/{data_source}_sequences_summary.csv"` file stored the name of the sequences file (which included the site_id) as well as the # of sequences stored in that file (and therefore the # of sequences associated with that lake). At this point, neither the mapping of which sequences belong to which site nor the record of how many sequences each lake has (which may change with this check for nans among the features, below) is being retained, right? Is that information that would be needed at any point down the road, or is it not relevant past this point because the model will be trained as a global model?

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # Remove lakes with 0 sequences
+    nonzero_sequences = sequences_summary[sequences_summary.num_sequences>0]
+    num_lakes = len(nonzero_sequences)
+
+    # Split lake indices into train and test
+    train_lakes, test_lakes = split_indices(num_lakes, [train_frac, test_frac], seed=seed)
+
+    lake_sequences_files_train = nonzero_sequences.sequences_file.iloc[train_lakes]
+    lake_sequences_files_test = nonzero_sequences.sequences_file.iloc[test_lakes]
```
hcorson-dosch, line 119: I might just be missing this, but are we retaining this information about which files fall into each group anywhere? I seems like it would be good to have a record of which lakes will be used for training and which will be used for testing.  I could see storing either these file names, or extracting the site_ids from the file names and storing those.

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    :param n_depths: Number of discrete depths at which LSTM will
+    output temperatures
+    :returns: numpy array of sequences
+
+    """
+    sequences = np.concatenate([np.load(sfn) for sfn in lake_sequence_files])
+    # remove inputs with nans
+    # shape of sequences is (# sequences, sequence_length, # depths + # features)
+    # slice from n_depths forward to index features
+    nan_inputs = np.any(np.isnan(sequences[:, :, n_depths:]), axis=(1,2))
```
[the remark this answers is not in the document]
AndyMcAliley: Your understanding is spot on. The reason for this step is that the neural network will fail if any inputs to the neural network are NaN. We're expecting NaNs among observations, and can handle those without a problem. However, if NaN values are input (either dynamic or static), then NaN values will be output, and training won't happen. There are some NaN values in the MNTOHA dataset in clarity and ice flags for a few lakes. So, this step is needed to obtain usable training and testing data.

2_process/src/training_data.py
AndyMcAliley: Good question. Since we're training a global model, I don't think we'll need that info down the road. If we do, I think we could add columns to the summary file without too much trouble.

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    # 2. Read lake-specific sequences from .npy files and concatenate
+
+    unscaled_train_data = read_sequences(lake_sequences_files_train, n_depths)
+    unscaled_test_data = read_sequences(lake_sequences_files_test, n_depths)
+
+    # 3. Standardize using training data
+
+    feature_means, feature_stds = get_mean_std(
+        unscaled_train_data,
+        axis=(0,1)
```
hcorson-dosch, line 130: Is there any reason you wouldn't always be calculating the means and standard deviations along these axes? I'm guessing maybe, since you specifically provided it as a kwarg, but if not I'd move this into the function. If it's kept here, I'd add a note to the function documentation that the passed kwargs include the axes along which to calculate the mean and standard deviation.

2_process/src/training_data.py
AndyMcAliley: You're right! We'll want to know which lakes were used for training and testing. I think another summary file would be useful for saving that info.

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    )
+
+    # Last dimension of data is features
```
hcorson-dosch, line 133: I'm confused by this comment. This and the name of the aggregated arrays (`feature_means` and `feature_stds` make it seem like only the features are being standardized, but aren't observations also being standardized here?

I may well just be missing something 😅

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    train_data = (unscaled_train_data - feature_means)/feature_stds
+    test_data = (unscaled_test_data - feature_means)/feature_stds
```
hcorson-dosch, line 135: Just confirming - we're standardizing the `test_data` with the `feature_means` and `feature_stds` calculated from the `train_data` because we want them to be consistently standardized, right?

2_process/src/training_data.py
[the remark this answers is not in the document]
hcorson-dosch: Ok great - thanks for confirming. That makes complete sense.

2_process/src/training_data.py
hcorson-dosch: Ah, I see. Yes I was thinking of the observations also as inputs to the training process. But another ML person would likely have the same interpretation as you do. I think 'input features' sounds like a good clarification.

hcorson-dosch (commented):
Okay I think I'm done reviewing for now, @AndyMcAliley . Overall this looks great. I found your comments and documentation very thorough and helpful. I had a few Qs and notes about some additional info that could be provided in comments, but I think the only edits I'm requesting are that you add that summary file to document which lakes fall into the train and test splits, and change the name of the `feature_means` and `feature_stds` stored arrays to indicate that they were calculated from the training data.

2_process/src/training_data.py
AndyMcAliley: Probably there wouldn't be a reason to calculate means and standard deviations along other axes. In fact, since this function is so small, I could get rid of it and move the 3 lines of code directly into `get_train_test_data`.

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+    lake_sequences_files_test = nonzero_sequences.sequences_file.iloc[test_lakes]
+
+    # 2. Read lake-specific sequences from .npy files and concatenate
+
+    unscaled_train_data = read_sequences(lake_sequences_files_train, n_depths)
+    unscaled_test_data = read_sequences(lake_sequences_files_test, n_depths)
+
+    # 3. Standardize using training data
+
+    feature_means, feature_stds = get_mean_std(
```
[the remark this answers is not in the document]
AndyMcAliley: Great point. I'm thinking I'll rename to `train_data_means` and `train_data_stds`.

2_process/src/training_data.py
AndyMcAliley: Yes, both input features and output targets are standardized here! I'll reword the comment for clarity.

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+        snakemake.params['seed']
+    )
+    # save data to npz along with metadata
+    npz_file_train = snakemake.output['train_file']
+    npz_file_test = snakemake.output['test_file']
+    # **snakemake.params saves all snakemake params to npz, using their names
+    # as keywords
+    np.savez(npz_file_train,
+             data=train_data,
+             feature_means=feature_means,
```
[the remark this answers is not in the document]
AndyMcAliley: Will do!

2_process/src/training_data.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # 3. Standardize using training data
+
+    feature_means, feature_stds = get_mean_std(
+        unscaled_train_data,
+        axis=(0,1)
+    )
+
+    # Last dimension of data is features
+    train_data = (unscaled_train_data - feature_means)/feature_stds
```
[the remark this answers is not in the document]
AndyMcAliley: I'll clarify that in the comments.

AndyMcAliley:
The commits above address the review comments. In the process, some functions were reorganized:

1. `get_mean_std` was removed. Instead, means and standard deviations are computed inside `get_train_test_data`.
2. `get_train_test_sequences_files` was added. This function assigns lakes to the training set and the test set. This was originally Step 1 of `get_train_test_data`. Having it as a separate function makes it easier to save a file that logs which lakes are in the training set and which lakes are in the test set.
3. A `main` function was added to contain the code that was in the body of the script.

Also, in 2_process.smk, the full config dictionary is passed as a parameter instead of passing some values from the dictionary. This makes it easier to save the entire config dictionary to `train.npz` and `test.npz`, including settings that were used in earlier parts of the pipeline.

hcorson-dosch (approved):
Your changes all look good to me, Andy!

PR #25, opened 2022-04-05
Add LSTM Model classes

### LINKED ISSUE #19: Add LSTM and EA-LSTM classes
Add classes that implement the LSTM and the EA-LSTM. Implementations can be found in the [code accompanying Kratzert et al., 2019](https://github.com/kratzert/ealstm_regional_modeling) and [Willard et al., 2022](https://github.com/jdwillard19/lake_conus_surface_temp_2021). Both of those implementations have only a single output of the LSTM, so we'll need to modify the classes to accommodate multiple outputs (temperatures at multiple depths).

AndyMcAliley:
This PR adds classes that define the LSTM and the EA-LSTM. Closes #19.

The first commit copies the model classes directly from the code supporting Kratzert et al., 2019. The second commit makes the necessary modifications to the classes, so the diff for that commit should be useful for reviewing.

## Three classes
There are three classes defined: `LSTM`, `EALSTM`, and `Model`. The `Model` class is a wrapper for both the `LSTM` and the `EALSTM` classes. It's a convenience class that allows a user to use either an LSTM or a EA-LSTM based on the boolean inputs `concat_static` and `no_static`.

## My modifications to the classes
I modified the classes to achieve two goals that the original code wasn't set up to achieve.

1. Provide predictions at every time step. The original code only produces a prediction for the final time step.
2. Allow multiple output values at every time step. This is needed because we want to predict temperatures at multiple depths at once.

## Additional context

Right now there's no code that instantiates objects from these classes. I'll submit another PR that trains the model, so there will be an opportunity to see the `Model` class in action then. Some feedback could wait for that PR. I can provide (messy!) code to play with the model classes if desired.

These classes were originally found in the repository https://github.com/kratzert/ealstm_regional_modeling/ and modified. They were written in support of

Kratzert, F., Klotz, D., Shalev, G., Klambauer, G., Hochreiter, S., and Nearing, G.: Towards learning 
universal, regional, and local hydrological behaviors via machine learning applied to large-sample 
datasets, Hydrol. Earth Syst. Sci., 23, 5089–5110, https://doi.org/10.5194/hess-23-5089-2019, 2019. 

Some modifications are similar to those found in https://github.com/jdwillard19/lake_conus_surface_temp_2021. Those modifications support the publication

Willard, J. D., Read, J. S., Topp, S., Hansen, G. J., & Kumar, V. Daily surface temperatures for 185,549 lakes in the conterminous United States estimated using deep learning (1980–2020). Limnology and Oceanography Letters, https://doi.org/10.1002/lol2.10249, 2022.

### DIFF
--- a/3_train/src/models.py
+++ b/3_train/src/models.py
@@ -274,6 +274,7 @@ class Model(nn.Module):
                  input_size_dyn: int,
                  input_size_stat: int,
                  hidden_size: int,
+                 output_size: int,
                  initial_forget_bias: int = 5,
                  dropout: float = 0.0,
                  concat_static: bool = False,
@@ -288,6 +289,8 @@ class Model(nn.Module):
             Number of static input features (used in the EA-LSTM input gate).
         hidden_size: int
             Number of LSTM cells/hidden units.
+        output_size: int
+            Number of LSTM output units.
         initial_forget_bias: int
             Value of the initial forget gate bias. (default: 5)
         dropout: float
@@ -317,7 +320,7 @@ class Model(nn.Module):
                                initial_forget_bias=initial_forget_bias)
 
         self.dropout = nn.Dropout(p=dropout)
-        self.fc = nn.Linear(hidden_size, 1)
+        self.fc = nn.Linear(hidden_size, output_size)
 
     def forward(self, x_d: torch.Tensor, x_s: torch.Tensor = None) \
             -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
@@ -343,6 +346,6 @@ class Model(nn.Module):
             h_n, c_n = self.lstm(x_d)
         else:
             h_n, c_n = self.lstm(x_d, x_s)
-        last_h = self.dropout(h_n[:, -1, :])
-        out = self.fc(last_h)
+        h_n = self.dropout(h_n)
+        out = self.fc(h_n)
         return out, h_n, c_n
--- a/3_train/src/models.py
+++ b/3_train/src/models.py
@@ -1,0 +1,348 @@
+# Model classes
+
+## From https://github.com/kratzert/ealstm_regional_modeling/blob/64a446e9012ecd601e0a9680246d3bbf3f002f6d/papercode/lstm.py
+
+"""
+This file is part of the accompanying code to our manuscript:
+
+Kratzert, F., Klotz, D., Shalev, G., Klambauer, G., Hochreiter, S., Nearing, G., "Benchmarking
+a Catchment-Aware Long Short-Term Memory Network (LSTM) for Large-Scale Hydrological Modeling".
+submitted to Hydrol. Earth Syst. Sci. Discussions (2019)
+
+You should have received a copy of the Apache-2.0 license along with the code. If not,
+see <https://opensource.org/licenses/Apache-2.0>
+"""
+
+from typing import Tuple
+
+import torch
+import torch.nn as nn
+
+
+class LSTM(nn.Module):
+    """Implementation of the standard LSTM.
+
+    TODO: Include ref and LaTeX equations
+
+    Parameters
+    ----------
+    input_size : int
+        Number of input features
+    hidden_size : int
+        Number of hidden/memory cells.
+    batch_first : bool, optional
+        If True, expects the batch inputs to be of shape [batch, seq, features] otherwise, the
+        shape has to be [seq, batch, features], by default True.
+    initial_forget_bias : int, optional
+        Value of the initial forget gate bias, by default 0
+    """
+
+    def __init__(self,
+                 input_size: int,
+                 hidden_size: int,
+                 batch_first: bool = True,
+                 initial_forget_bias: int = 0):
+        super(LSTM, self).__init__()
+
+        self.input_size = input_size
+        self.hidden_size = hidden_size
+        self.batch_first = batch_first
+        self.initial_forget_bias = initial_forget_bias
+
+        # create tensors of learnable parameters
+        self.weight_ih = nn.Parameter(torch.FloatTensor(input_size, 4 * hidden_size))
+        self.weight_hh = nn.Parameter(torch.FloatTensor(hidden_size, 4 * hidden_size))
+        self.bias = nn.Parameter(torch.FloatTensor(4 * hidden_size))
+
+        # initialize parameters
+        self.reset_parameters()
+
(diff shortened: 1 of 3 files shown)


3_train/src/models.py
```diff
@@ -0,0 +1,351 @@
...
+
+        weight_hh_data = torch.eye(self.hidden_size)
+        weight_hh_data = weight_hh_data.repeat(1, 4)
+        self.weight_hh.data = weight_hh_data
+        nn.init.constant_(self.bias.data, val=0)
+
+        if self.initial_forget_bias != 0:
+            self.bias.data[:self.hidden_size] = self.initial_forget_bias
+
+    def forward(self, x: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
jdiaz4302, line 72: Flow and operations check out 👍🏼

3_train/src/models.py
```diff
@@ -0,0 +1,351 @@
...
+        weight_hh_data = weight_hh_data.repeat(1, 3)
+        self.weight_hh.data = weight_hh_data
+
+        nn.init.constant_(self.bias.data, val=0)
+        nn.init.constant_(self.bias_s.data, val=0)
+
+        if self.initial_forget_bias != 0:
+            self.bias.data[:self.hidden_size] = self.initial_forget_bias
+
+    def forward(self, x_d: torch.Tensor, x_s: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
```
jdiaz4302, line 203: Flow and operations check out 👍🏼 

The math is a strange idea at first, but I like it and think it makes sense to use a fixed input gate to modulate the candidate for short term memory. I was originally assuming the static features would be used only in a fully connected layer on top of the LSTM to sort of nudge the dynamic outputs, but I like this approach of them being more tightly coupled with the temporal dynamics.

3_train/src/models.py
```diff
@@ -0,0 +1,351 @@
...
+        super(Model, self).__init__()
+        self.input_size_dyn = input_size_dyn
+        self.input_size_stat = input_size_stat
+        self.hidden_size = hidden_size
+        self.initial_forget_bias = initial_forget_bias
+        self.dropout_rate = dropout
+        self.concat_static = concat_static
+        self.no_static = no_static
+
+        if self.concat_static or self.no_static:
```
jdiaz4302, line 312: I like this, and more broadly, the class is a good minimal way to add the fully connected layer

3_train/src/models.py
```diff
@@ -0,0 +1,351 @@
...
+            f, i, o, g = gates.chunk(4, 1)
+
+            c_1 = torch.sigmoid(f) * c_0 + torch.sigmoid(i) * torch.tanh(g)
+            h_1 = torch.sigmoid(o) * torch.tanh(c_1)
+
+            # store intermediate hidden/cell state in list
+            h_n.append(h_1)
+            c_n.append(c_1)
+
+            h_x = (h_1, c_1)
```
jdiaz4302, line 118: The notation of putting h and c into an object named h could become misleading for newcomers or when you're debugging

3_train/src/models.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # perform forward steps over input sequence
+        for t in range(seq_len):
+            h_0, c_0 = h_x
+
+            # calculate gates
+            gates = (torch.addmm(bias_batch, h_0, self.weight_hh) + torch.mm(x[t], self.weight_ih))
+            f, i, o, g = gates.chunk(4, 1)
+
+            c_1 = torch.sigmoid(f) * c_0 + torch.sigmoid(i) * torch.tanh(g)
```
jdiaz4302, line 111: the "_1" notation is a little strange, I was tempted to suggest that it be "_t" but I guess it makes since if you think of it relative to the last data point since you're always starting with "_0" as well (i.e., it's not truly the 0th, it's the last one)

jdiaz4302:
Left specific comments. Notation comments are more or less take/leave, I don't feel very strongly about them but I gave my thoughts. I followed the matrix shapes and calculations, everything corresponds to the equations.

Curious about the bigger weight initialization comment that I left, but I'm not too concerned about it either - if it works in practice, it's fine; it might be worth looking at if you have trouble training the models at some point though

3_train/src/models.py
AndyMcAliley: Yes, it's an intuitively pleasing concept after thinking it through!

3_train/src/models.py
AndyMcAliley: It makes it easy to compare the EA-LSTM and a vanilla LSTM with static and dynamic attributes concatenated together.

3_train/src/models.py
AndyMcAliley: Agreed on all counts! Again, I'll leave the author's notation as is. In the past I've used `x_i` for the current, `x_p` for the next (i plus one), and `x_m` for the previous (i minus one), but wasn't totally happy with that system because I don't think it's immediately apparent to a new reader what it means, and because the p in `x_p` could be misinterpreted as "previous".

3_train/src/models.py
AndyMcAliley: I'm not sure what the reasoning behind naming this variable `h_x` was. I think I'll leave the author's variable names as they are.

3_train/src/models.py
```diff
@@ -0,0 +1,351 @@
...
+
+        # create tensors of learnable parameters
+        self.weight_ih = nn.Parameter(torch.FloatTensor(input_size, 4 * hidden_size))
+        self.weight_hh = nn.Parameter(torch.FloatTensor(hidden_size, 4 * hidden_size))
+        self.bias = nn.Parameter(torch.FloatTensor(4 * hidden_size))
+
+        # initialize parameters
+        self.reset_parameters()
+
+    def reset_parameters(self):
```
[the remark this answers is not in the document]
AndyMcAliley: It's a good question! I don't see anything in Kratzert et al., 2019 about why they made that choice. Here's my best guess. Intuitively, if the weights are identity matrices, then the initial model has:

1. Stability: Seems like `h` won't change as much from one time step to the next, compared to randomly initialized weights.
2. Independence: each element of `h` at the next time step depends only upon the corresponding element of `h` at the previous time step (and the inputs `x`).

Seems like a way to encourage stable outputs during early training epochs, and thus faster training and more dependable convergence.

PR #26, opened 2022-04-14
Train model

### LINKED ISSUE #20: Train EA-LSTM model using MNTOHA data
Train an EA-LSTM using MNTOHA data. The code for the training loop of the model should be fairly short and straightforward. We'll track loss over epochs. After training, we'll save the parameters of the trained model and the metrics tracked during training.

AndyMcAliley:
This PR allows an LSTM or an EA-LSTM to be trained to predict lake temperatures at multiple depths using MNTOHA data. Closes #20.

The structure of the functions is patterned after [this tutorial](https://pytorch.org/tutorials/beginner/nn_tutorial.html#create-fit-and-get-data).

To execute the full pipeline and train the model for 3 epochs, run
```
conda update -n base conda
conda env create -f environment.yaml
conda activate ltls
snakemake --cores all 3_train/out/mntoha/0/lstm.pt
```
NOTE: This will require about half an hour and 7 GB of disk space. The pipeline downloads MNTOHA data from ScienceBase, forms training data, and trains an EA-LSTM for 3 epochs as a test.

DONE:
- Save losses at every epoch
- Average loss weighted by the number of non-NaN observations
- Save settings and other metadata alongside model weights

TODO in later PRs, in order of importance (time allowing):
- Implement early stopping
- Allow pre-trained model to be used
- Split validation data apart by lake, not sequence
- Allow different weight initializations
- Allow different optimizers
- Implement cross-validation
- Save space by saving only one copy of static attributes to training data .npz

### DIFF
--- a/3_train/src/train.py
+++ b/3_train/src/train.py
@@ -246,3 +246,50 @@ def loss_batch(model, loss_func, x_d, x_s, y, opt=None):
     # return loss.item(), len(y)
     return loss.item(), torch.sum(loss_idx)
 
+
+def fit(epochs, model, loss_func, opt, train_dl, valid_dl):
+    """
+    Train the model, and compute training and validation losses for each epoch
+
+    Patterned after https://pytorch.org/tutorials/beginner/nn_tutorial.html#create-fit-and-get-data
+
+    :param epochs: Maximum number of training epochs
+    :param model: PyTorch model inheriting nn.Module()
+    :param loss_func: PyTorch loss function, e.g., nn.MSELoss()
+    :param opt: Optimizer to use to update model parameters
+    :param train_dl: PyTorch DataLoader with training data
+    :param valid_dl: PyTorch DataLoader with validation data
+
+    """
+
+    # Count number of observations in training and validation sets
+    n_train = 0
+    for x_d, x_s, y in train_dl:
+        n_train += torch.sum(torch.isfinite(y))
+    n_valid = 0
+    for x_d, x_s, y in valid_dl:
+        n_valid += torch.sum(torch.isfinite(y))
+
+    train_losses = []
+    valid_losses = []
+    print('Epoch: train loss, validate loss')
+    for epoch in range(epochs):  # loop over the dataset multiple times
+        model.train()
+        train_loss = 0.0
+        # data is ordered as [dynamic inputs, static inputs, labels]
+        for x_d, x_s, y in train_dl:
+            batch_loss, batch_count = loss_batch(model, loss_func, x_d, x_s, y, opt)
+
+            # Track this epoch's loss
+            train_loss += batch_loss * batch_count/n_train
+
+        model.eval()
+        with torch.no_grad():
+            losses, nums = zip(
+                *[loss_batch(model, loss_func, x_d, x_s, y) for x_d, x_s, y in valid_dl]
+            )
+        val_loss = np.sum(np.multiply(losses, nums)) / np.sum(nums)
+        print(f'{epoch}: {train_loss}, {val_loss}')
+        train_losses.append(train_loss)
+        valid_losses.append(val_loss)
+
--- a/3_train/src/train.py
+++ b/3_train/src/train.py
@@ -73,7 +73,7 @@ def split(dataset, fraction, seed):
 
     :param dataset: A torch.utils.data.Dataset to be split
     :param fraction: The fraction of data samples to put in the first subset
-    :param seed: Seed for random number generator
+    :param seed: Integer seed for random number generator
     :returns: Tuple of two datasets for the two subsets
 
     """
@@ -124,7 +124,7 @@ def get_data(npz_filepath,
     :param depths_use: List of depth values to include
     :param batch_size: Number of elements in each training batch
     :param valid_frac: The fraction of data samples to put into the validation set
-    :param seed: Seed for random number generator
+    :param seed: Integer seed for random number generator
 
     """
 
@@ -184,7 +184,7 @@ def get_model(
     :param n_static: Number of static input features
     :param hidden_size: Number of elements in hidden state and cell state
     :param initial_forget_bias: Value of the initial forget gate bias
-    :param dropout: Dropout probability, from 0 to 1
+    :param dropout: Dropout probability, from 0 to 1 (0 = don't use dropout)
     :param concat_static: If True, uses standard LSTM. Otherwise, uses EA-LSTM
     :param learning_rate: Learning rate for optimizer
     :returns: Tuple of model and optimizer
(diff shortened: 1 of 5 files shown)


3_train/src/train.py
```diff
@@ -0,0 +1,465 @@
...
+    19069
+    >>> # dynamic features, static features, and temperatures at depths
+    >>> # for first sequence
+    >>> for e in dataset[0]:
+    >>>     print(e.shape)
+    >>>
+    torch.Size([400, 9])
+    torch.Size([400, 4])
+    torch.Size([400, 49])
+    """
```
jdiaz4302, line 45: really good documentation

3_train/src/train.py
```diff
@@ -0,0 +1,465 @@
...
+    :param seed: Integer seed for random number generator
+    :returns: Tuple of two datasets for the two subsets
+
+    """
+    # split into two subsets
+    n_total = len(dataset)
+    n_1 = int(round(n_total * fraction))
+    n_2 = n_total - n_1
+    subset_1, subset_2 = random_split(
+        dataset, [n_1, n_2], generator=torch.Generator().manual_seed(seed))
```
jdiaz4302, line 87: I like this setting the seed as an argument, haven't seen that before (I usually just set it at the top of the script, but that is dependent on the same order and number of random operations)

3_train/src/train.py
```diff
@@ -0,0 +1,465 @@
...
+    n_depths,
+    n_dynamic,
+    n_static,
+    hidden_size,
+    initial_forget_bias,
+    dropout,
+    concat_static
+):
+    """
+    Create LSTM torch model
```
jdiaz4302, line 178: Might be good to have the documentation mention that this is LSTM _and_ EA LSTM compatible (because that Model class can be both)

3_train/src/train.py
```diff
@@ -0,0 +1,465 @@
...
+    loss_idx = torch.isfinite(y)
+    loss = loss_func(pred[loss_idx], y[loss_idx])
+
+    if opt is not None:
+        loss.backward()
+        opt.step()
+        opt.zero_grad()
+
+    # return loss, number of finite values in y
+    return loss.item(), torch.sum(loss_idx).item()
```
jdiaz4302, line 246: Do you ever have a batch that is all `nan` y observations? If so, has this been stress tested for that case (i.e. does line 238 error out with an empty input)

One of the more succinct loss functions for handling `nan` 🚀 . I like including that the training operations are in it too 😄 (241-243)

3_train/src/train.py
```diff
@@ -0,0 +1,465 @@
...
+        os.makedirs(directory)
+
+    if not overwrite:
+        # Append unique suffix to filename
+        root, extension = os.path.splitext(filepath)
+        suffix = 0
+        while os.path.exists(filepath):
+            suffix += 1
+            filepath = f'{root}_{suffix}{extension}'
+    torch.save(model.state_dict(), filepath)
```
jdiaz4302, line 337: I like all of this - robust

3_train/src/train.py
```diff
@@ -0,0 +1,465 @@
...
+        concat_static         boolean, Add static to dynamic and use standard LSTM, or not
+        dropout               float, Dropout probability, from 0 to 1 (0 = don't use dropout)
+        initial_forget_bias   float, Value of the initial forget gate bias
+        hidden_size           integer, Number of elements in hidden state and cell state
+        static_features_use   list of strings, Names of static features for model to use
+        dynamic_features_use  list of strings, Names of dynamic features for model to use
+        depths_use            list of floats, depth values of temperatures for model to use
+        seed                  integer, Seed for the random number generator
+        valid_frac            float, Fraction of examples to put into the validation set
+        batch_size            integer, Number of examples per training batch
```
jdiaz4302, line 407: again, really nice documentation

jdiaz4302:
Your TO DOs look good too!

`Allow different optimizers`

For what it's worth, I've generally found this one to not matter (default Adam is great). Your experience may be different though, and I'd like to see if it ends up being so

jdiaz4302:
> NOTE: This will require about half an hour and 7 GB of disk space. 

This seems fine. Do you know what the computation time looks like with the number of epochs you'll actually use?

AndyMcAliley:
> Do you know what the computation time looks like with the number of epochs you'll actually use?

I don't, but I'm expecting something on the order of days. It'll depend in large part on the size of the hidden state. Right now I'm using 20 for testing, but will increase that. Of the 30 minutes it takes right now, only a few are actually for training the model. Mostly the 30 minutes are spent downloading and forming the training data.

AndyMcAliley:
> default Adam is great

Agreed! I've had the same experience. This TODO item is lower priority.

3_train/src/train.py
AndyMcAliley: Thanks! This loss function is a hybrid of [this PyTorch tutorial](https://pytorch.org/tutorials/beginner/nn_tutorial.html#create-fit-and-get-data) and [Jared Willard's surface temp code](https://github.com/jdwillard19/lake_conus_surface_temp_2021/blob/88334091dec71ae43fe4256603d65045141936b5/src/train/EALSTM_final_model.py#L580-L586). Good point about a batch of all `nan` observations. Every sequence in the training and testing set should include at least one observation. It would be easy to check for this and raise an error, in case of bug.

3_train/src/train.py
AndyMcAliley: I set the seed on line 411 too, so this approach might be a bit redundant. The nice thing is that the train/validation split will be predictable, apart from any other changes to the workflow.

jdiaz4302: I don't think it's redundant at all! As a general take-away for myself, I was trying to say that it's a good idea because it's not fragile to some new operation (you may insert later in a generic script) occurring before it and eating up its random seed value/iteration.

Ex of why your way is good

SeedAsArg

3_train/src/train.py
jdiaz4302: For sure, I was just thinking it would be good to know before you get (e.g.,) multiple hours/days into training. Could use a simply `try/catch` set up to handle the error

3_train/src/train.py
AndyMcAliley: Nicely demonstrated!

3_train/src/train.py
AndyMcAliley: Yes - added! I clarified that in a couple of other comments, too.

jdiaz4302:
Looking good, I think that's all I have!

I wouldn't be able to set it up and run it anytime soon, but I trust that you've got it working

3_train/src/train.py
AndyMcAliley: Yes, `NaN`s sneaking into the loss are the pits! I tested `loss_batch` with all `NaN` labels, and it silently returned `NaN` as the loss value. So, I added a check that raises a `ValueError`. I also added a check to make sure the return value isn't `NaN`.

jdiaz4302: Cool!

I don't know what your batch size or `NaN` percentage is like, but I'm thinking that this would still crash your whole model run if it happens upon a batch of all `NaN` (effectively undo'ing hours/days of work). This is more likely to be the case if you have a small batch size and lots of `NaN` (maybe this isn't the case though and that's why the next part isn't necessary, so feel free to disregard if I'm trying to solve a problem that doesn't exist)

I think a preferable approach would be to skip past a batch if it doesn't have any real observations to train with - e.g., with partial pseudocode

```
for batch_x, batch_y in data_loader:
    if sum(torch.isfinite(batch_y)) == 0:
        continue
    else:
        compute loss
```

This would prevent the loss from being `NaN` and not error out the pipeline

AndyMcAliley: I see your point. That would crash the training, potentially after running for a few days. However, in the case of this project, that crash would indicate that there's probably a bug in `2_process` that should be fixed. There shouldn't be a single training (or testing) sequence that consists of all `NaN` observations, much less a batch of them. During `2_process`, only sequences that include at least one finite temperature observation are included in the training and testing sets.

AndyMcAliley: I lean toward leaving the code as is. If that error is never raised - great. If the error is reached, the results shouldn't be usable anyway since there's an unexpected problem with the training data that needs to be addressed - something's wrong with the way we're forming the training data in that case.

PR #27, opened 2022-04-21
Log settings and other metadata after training

### LINKED ISSUE #18: Save results for different hyperparameters, etc. to different places to avoid overwriting older results
Following up on [this discussion](https://github.com/USGS-R/lake-temperature-lstm-static/discussions/14#discussioncomment-2424230), we want to save multiple sets of results that correspond to different choices of hyperparameters, processing decisions, etc. This issue is also linked to the question of how to log and track those choices alongside results.

AndyMcAliley:
Save configuration settings, details about training, and preliminary performance metrics after training a model. Summarize all of that metadata for all trained models in one csv file. Closes #18.

This PR provides a way to summarize all the trained models so far in a csv. 
`snakemake -c1 3_train/out/mntoha/summary.csv`
There is also an option to summarize all the models in a single run only. This is useful when a run is a hyperparameter search that produced many models.
`snakemake -c1 3_train/out/mntoha/tune_hidden_size/summary.csv`
NOTE: the above commands won't work if there are no npz files associated with trained models to summarize. If it would help, I can provide a series of commands that would quickly train a couple of models and then summarize them.

As an example, the structure of `3_train/out` might look like this:
```
3_train/
| out/
| | mntoha/
| | | dynamic_inputs_1/
| | | | a_metadata.npz
| | | | a_weights.pt
| | | | process_config.yaml
| | | | train_config.yaml
| | | dynamic_inputs_2/
| | | | a_metadata.npz
| | | | a_weights.pt
| | | | process_config.yaml
| | | | train_config.yaml
| | | tune_hidden_size_and_dropout/
| | | | hidden_size-50_dropout-0.0_metadata.npz
| | | | hidden_size-50_dropout-0.0_weights.pt
| | | | hidden_size-100_dropout-0.0_metadata.npz
| | | | hidden_size-100_dropout-0.0_weights.pt
| | | | hidden_size-50_dropout-0.2_metadata.npz
| | | | hidden_size-50_dropout-0.2_weights.pt
| | | | hidden_size-100_dropout-0.2_metadata.npz
| | | | hidden_size-100_dropout-0.2_weights.pt
| | | | process_config.yaml
| | | | train_config.yaml
| | | summary.csv
| | full_footprint
| | | dynamic_inputs_1/
| | | | a_metadata.npz
| | | | a_weights.pt
| | | | process_config.yaml
| | | | train_config.yaml
| | | summary.csv
```
Here there are two data sources: `mntoha` and `full_footprint`. `mntoha` has three runs with `run_id` values `dynamic_inputs_1`, `dynamic_inputs_2`, and `tune_hidden_size_and_dropout`. The first two runs have only one `model_id`: `a`. The third run did a grid search: two hyperparameters, searching over two values apiece. The third run has multiple models with different `model_id`s corresponding to the values tested during the grid search, like `hidden_size-50_dropout-0.0`, for example.

The PR also adds to the information that is saved after training. 

## Metadata saved:

- full_id (combination of run_id and model_id)
- sequence_length
- sequence_offset
- spinup_time
- depths
- train_frac
- test_frac
- depths_all (all discrete depths used during processing to interpolate temperature observations)
- dynamic_features_all
- static_features_all
- seed
- run_id (set in train_config.yaml and via the snakemake command; these two must agree)
- run_description
- overwrite (boolean of whether or not to overwrite existing results. Right now, ineffective)
- dynamic_features_use
- static_features_use
- depths_use (all discrete depths used during training)
- hidden_size
- concat_static (boolean of whether to use an EA-LSTM or regular LSTM)
- valid_frac
- k_folds
- batch_size
- initial_forget_bias
- dropout
- begin_loss_ind
- learning_rate
- loss_criterion
- early_stopping
- max_epochs
- model_id (set via the snakemake command. Going forward, this can be used to distinguish among models during a grid search)
- train_start_time
- train_end_time
- n_epochs_trained
- git_hash
- git_status (full message returns by running `git status` - useful to see if the repo was clean or not when the model was trained)

The metadata includes all the info from `process_config.yaml` and `train_config.yaml`. Those yaml files also get saved directly to the run folder as well. There's some redundancy there, but I see no harm in it.

## Note about pipeline structure

I decoupled the rule to summarize, `summarize_trained_models`, from the rest of the pipeline. My concern that running that rule might accidentally trigger re-training a model, so I gave that rule no inputs. Hope that makes sense!

# How to review this PR

After today's sprint meeting, I'm experimenting with this section! Let's see if it helps...

## Level of review requested

Since we want to get some preliminary models trained soon, the main things I'd like reviewed are:

1. Are there any obvious reasons the code won't work as expected?
2. Are there any "gotchas" lying in wait?
3. Is there any additional metadata that should be saved after training?
4. Are there changes that will make things significantly easier for us going forward?

I'm happy to clarify any questions about how the code should work.

If there are substantial edits that would be good-to-have but wouldn't be necessary at this point, feel free to suggest them and I'll spin them off into a new issue!

## Where in the code to focus

Basically, all new/edited snakemake or python code is fair game. There are some edits to docstrings that don't warrant close attention.

### DIFF
--- a/3_train.smk
+++ b/3_train.smk
@@ -1,10 +1,27 @@
 
 configfile: "3_train/train_config.yaml"
 
+# Save config files to output directory
+rule save_config:
+    input:
+        process_config = "2_process/process_config.yaml",
+        train_config = "3_train/train_config.yaml"
+    output:
+        process_config = "3_train/out/{data_source}/{run_id}/process_config.yaml",
+        train_config = "3_train/out/{data_source}/{run_id}/train_config.yaml"
+    shell:
+        """
+        mkdir -p 3_train/out/{wildcards.data_source}/{wildcards.run_id}
+        cp {input.process_config} {output.process_config}
+        cp {input.train_config} {output.train_config}
+        """
+
 # Train model
 rule train_model:
     input:
-        npz_filepath = "2_process/out/{data_source}/train.npz"
+        npz_filepath = "2_process/out/{data_source}/train.npz",
+        process_config = "3_train/out/{data_source}/{run_id}/process_config.yaml",
+        train_config = "3_train/out/{data_source}/{run_id}/train_config.yaml"
     output:
         weights_filepath = "3_train/out/{data_source}/{run_id}/lstm.pt",
         metadata_filepath = "3_train/out/{data_source}/{run_id}/metadata.npz"
--- a/3_train.smk
+++ b/3_train.smk
@@ -22,7 +22,8 @@ rule save_config:
 rule train_model:
     input:
         npz_filepath = "2_process/out/{data_source}/train.npz",
-        # Include configs as inputs in order to save them to the output folder
+        # Include configs as inputs in order to trigger save_config rule,
+        # and save configs to the output folder
         process_config = "3_train/out/{data_source}/{run_id}/process_config.yaml",
         train_config = "3_train/out/{data_source}/{run_id}/train_config.yaml"
     output:
--- a/3_train.smk
+++ b/3_train.smk
@@ -23,11 +23,14 @@ rule train_model:
         process_config = "3_train/out/{data_source}/{run_id}/process_config.yaml",
         train_config = "3_train/out/{data_source}/{run_id}/train_config.yaml"
     output:
-        weights_filepath = "3_train/out/{data_source}/{run_id}/lstm.pt",
-        metadata_filepath = "3_train/out/{data_source}/{run_id}/metadata.npz"
+        weights_filepath = "3_train/out/{data_source}/{run_id}/{model_id}_weights.pt",
+        metadata_filepath = "3_train/out/{data_source}/{run_id}/{model_id}_metadata.npz"
     params:
         config = config,
-        run_id = lambda wildcards: wildcards.run_id
+        run_id = lambda wildcards: wildcards.run_id,
+        model_id = lambda wildcards: wildcards.model_id
+    threads:
+        16
     script:
         "3_train/src/train.py"
 
--- a/3_train.smk
+++ b/3_train.smk
@@ -31,3 +31,11 @@ rule train_model:
     script:
         "3_train/src/train.py"
 
+# Summarize all training runs so far
+rule summarize_training_runs:
+    # No inputs because we don't want this rule to trigger the train_model rule
+    output: 
+        summary_filepath = "3_train/out/{data_source}/summary.csv"
+    script:
+        "3_train/src/summarize.py"
+
--- a/3_train.smk
+++ b/3_train.smk
@@ -6,10 +6,11 @@ rule train_model:
     input:
         npz_filepath = "2_process/out/{data_source}/train.npz"
     output:
-        weights_filepath = "3_train/out/{data_source}/{run}/lstm.pt",
-        metadata_filepath = "3_train/out/{data_source}/{run}/metadata.npz"
+        weights_filepath = "3_train/out/{data_source}/{run_id}/lstm.pt",
+        metadata_filepath = "3_train/out/{data_source}/{run_id}/metadata.npz"
     params:
-        config = config
+        config = config,
+        run_id = lambda wildcards: wildcards.run_id
     script:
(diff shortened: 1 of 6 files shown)


3_train.smk
```diff
@@ -1,15 +1,52 @@
...
+# Train a model and save weights and metadata
+# Example:
+# $ snakemake --cores all 3_train/out/mntoha/4/hidden_size-50_weights.pt
 rule train_model:
     input:
-        npz_filepath = "2_process/out/{data_source}/train.npz"
+        npz_filepath = "2_process/out/{data_source}/train.npz",
+        # Include configs as inputs in order to save them to the output folder
+        process_config = "3_train/out/{data_source}/{run_id}/process_config.yaml",
+        train_config = "3_train/out/{data_source}/{run_id}/train_config.yaml"
```
hcorson-dosch, line 27: Just to confirm that my understanding is correct -- these don't actually get passed to the `main` function in `3_train/src/train.py`, but including them here ensures that `snakemake` will build the `save_config` rule, correct?

3_train.smk
```diff
@@ -1,15 +1,52 @@
...
     output:
-        weights_filepath = "3_train/out/{data_source}/{run}/lstm.pt",
-        metadata_filepath = "3_train/out/{data_source}/{run}/metadata.npz"
+        weights_filepath = "3_train/out/{data_source}/{run_id}/{model_id}_weights.pt",
+        metadata_filepath = "3_train/out/{data_source}/{run_id}/{model_id}_metadata.npz"
     params:
-        config = config
+        config = config,
```
hcorson-dosch, line 32: So here, `config`, which refers to `"3_train/train_config.yaml"` is passed in `3_train/src/train.py` as an input to `save_metadata()` and therefore gets saved to the `metadata.npz` file in each run folder. That `.npz` file, in turn, is read in and saved through execution of the `summarize_trained_models` rule. However, the process config, `2_process/process_config.yaml`, while saved to the trained model output directory, is not ever written to a metadata `.npz` file and thus isn't included in the summary of trained models.  If that was your intent, then there's no issue -- I just wanted to  make note of that discrepancy since you are intentionally saving the process config to the model output directory alongside the train config.

hcorson-dosch: Disregard my previous comment - I remembered that the process config is actually [saved to the training data `.npz` files](https://github.com/USGS-R/lake-temperature-lstm-static/blob/main/2_process/src/training_data.py#L211-L215), so that   is captured and included in the metadata [here](https://github.com/AndyMcAliley/lake-temperature-lstm-static/blob/log-settings/3_train/src/train.py#L386-L392).

hcorson-dosch (approved):
Andy I think this looks like a good set up. The structure for the output directories make sense, the captured metadata seems helpful and complete, and you've documented the assumptions you're making when summarizing the metadata well. I think this is a great starting set up, and your upcoming training experiments and your subsequent evaluations will pretty quickly make it clear whether this metadata and set up is sufficient or if you want to make some tweaks.

3_train.smk
AndyMcAliley: Yes, exactly!

3_train/src/train.py
```diff
@@ -400,6 +429,7 @@ def main(npz_filepath, weights_filepath, metadata_filepath, config):
     :param npz_filepath: Name and path to .npz data file
     :param weights_filepath: Path and filename to save weights to
     :param metadata_filepath: Path and filename to save metadata to
+    :param run_id: ID of the current training run (experiment)
     :param config: Dictionary of configuration settings, including:
```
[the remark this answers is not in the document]
AndyMcAliley: Nice catch!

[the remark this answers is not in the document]
AndyMcAliley: Yes, good idea. Comments added to docstring.

3_train.smk
AndyMcAliley: Yes, that's right, both configs are combined and saved to the metadata `.npz`. I got worried there for a second...

PR #29, opened 2022-05-03
Tallgrass GPU bug fixes

### LINKED ISSUE #30: Train on a GPU
Let's get a model trained using a GPU on Tallgrass! There are some bugs that only pop up when we try this. Some are related to running on HPC resources, and some are specific to using GPUs.

AndyMcAliley:
This PR addresses issues that arose when trying to train a model using a GPU on Tallgrass. Closes #30. The changes can be organized into four categories:

1. Training data and validation data are now on the same device (cpu or gpu) as the model.
    - Shouldn't need much attention. This issue caused a show-stopping error, and now it doesn't. 
    - I could ask Jeremy how he's handled this after the temperature sprint ends.
2. Missing packages added to conda environment
    - The weird addition here is git. Conda will now install the git executable. This is so that the git status can be logged - turns out that compute nodes don't have git installed on them. I'm open to other options, but this does work!
3. Slurm scripts for both cpu and gpu have been written
    - I'd appreciate eyes here - I'm not fluent in slurm scripts, plus there's snakemake in the mix.
    - I adapted the command in [Jeff's blog](https://github.com/jsadler2/wdfn-blog/blob/snakemake-dl-experiments/content/snakemake-ml-experiments.md) for the snakemake part.
4. Miscellaneous minor changes
    - Add `setup.sh` to setup the environment easily
    - Git ignore the `log` folder where slurm output is directed
    - Add `flush=True` to Python `print()` commands to flush stdout and keep the logs up to date while the model is training. Otherwise it's hard to track the progress during training.

# How to review this PR

This one's a hodgepodge! Since it consists mainly of bug fixes that were needed to get a model out, it's been through a round of informal testing. Don't worry if there are parts that aren't in your wheelhouse - feel free to focus on the aspects of the PR that you can easily review, and let me know if there are aspects that you skim over. 

## Where in the code to focus

Feedback on the slurm scripts would be especially welcome!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -162,13 +162,13 @@ dmypy.json
 .Rproj.user
 
 # Specific files and directories to exclude
-1_fetch/tmp
-1_fetch/out
-2_process/log
-2_process/tmp
-2_process/out
-3_train/out
-log
+/1_fetch/tmp/
+/1_fetch/out/
+/2_process/log/
+/2_process/tmp/
+/2_process/out/
+/3_train/out/
+/log/
 
 # Specific files to include
 !/notebooks/Explore_MNTOHA_temperature_observations.html
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,8 @@
 *.patch
 *.pdf
 *.svg
+*.swp
+*.swo
 
 # Byte-compiled / optimized / DLL files
 __pycache__/
@@ -166,6 +168,7 @@ dmypy.json
 2_process/tmp
 2_process/out
 3_train/out
+log
 
 # Specific files to include
 !/notebooks/Explore_MNTOHA_temperature_observations.html
--- a/3_train/src/train.py
+++ b/3_train/src/train.py
@@ -192,12 +192,6 @@ def get_model(
     :returns: LSTM or EA-LSTM torch model
 
     """
-    if torch.cuda.is_available():
-        device = "cuda" 
-    else:
-        device = "cpu"
-    print(f"Using {device} device")
-
     model = Model(
         input_size_dyn=n_dynamic,
         input_size_stat=n_static,
@@ -206,7 +200,7 @@ def get_model(
         initial_forget_bias=initial_forget_bias,
         dropout=dropout,
         concat_static=concat_static
-    ).to(device)
+    )
 
     return model
 
@@ -257,7 +251,7 @@ def loss_batch(model, loss_func, x_d, x_s, y, opt=None):
     return loss.item(), torch.sum(loss_idx).item()
 
 
-def fit(epochs, model, loss_func, opt, train_dl, valid_dl, verbose=False):
+def fit(epochs, model, loss_func, opt, train_dl, valid_dl, device, verbose=False):
     """
     Train the model, and compute training and validation losses for each epoch
     
@@ -269,6 +263,7 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, verbose=False):
     :param opt: Optimizer to use to update model parameters
     :param train_dl: PyTorch DataLoader with training data
     :param valid_dl: PyTorch DataLoader with validation data
+    :param device: Device that pytorch is running on (cpu or gpu)
     :param verbose: Print more messages during execution
         (Default value = False)
     :returns: Tuple of arrays of training and validation losses for each epoch
@@ -298,6 +293,8 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, verbose=False):
         # Data is ordered as [dynamic inputs, static inputs, labels]
         verbose_print('Train loss by batch')
         for x_d, x_s, y in train_dl:
+            # Transfer to device
+            x_d, x_s, y = x_d.to(device), x_s.to(device), y.to(device)
             batch_loss, batch_count = loss_batch(model, loss_func, x_d, x_s, y, opt)
 
             # Weight the batch loss based on number of observations in each batch
@@ -310,6 +307,8 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, verbose=False):
         verbose_print('Validation loss by batch')
         with torch.no_grad():
             for x_d, x_s, y in valid_dl:
+                # Transfer to device
+                x_d, x_s, y = x_d.to(device), x_s.to(device), y.to(device)
                 batch_loss, batch_count = loss_batch(model, loss_func, x_d, x_s, y)
 
                 # Weight the batch loss based on number of observations in each batch
@@ -475,6 +474,16 @@ def main(npz_filepath, weights_filepath, metadata_filepath, run_id, model_id, co
     n_dynamic = len(config['dynamic_features_use'])
     n_static = len(config['static_features_use'])
 
+    if torch.cuda.is_available():
+        device = "cuda" 
+    else:
+        device = "cpu"
+    print(f"Using {device} device")
+
+    print(f"Number of threads set by user: {torch.get_num_threads()}")
+    print(f"Number of GPUs: {torch.cuda.device_count()}")
+    print(f"Number of CPUs: {os.cpu_count()}")
+
     # Create model
     model = get_model(n_depths,
                       n_dynamic,
@@ -482,7 +491,7 @@ def main(npz_filepath, weights_filepath, metadata_filepath, run_id, model_id, co
                       config['hidden_size'],
                       config['initial_forget_bias'],
                       config['dropout'],
-                      config['concat_static'])
+                      config['concat_static']).to(device)
 
     # Create optimizer
     optimizer = Adam(model.parameters(), lr=config['learning_rate'])
@@ -494,7 +503,7 @@ def main(npz_filepath, weights_filepath, metadata_filepath, run_id, model_id, co
 
     # Training loop
     train_start_time = str(datetime.now())
-    train_losses, valid_losses = fit(config['max_epochs'], model, loss_func, optimizer, train_data_loader, valid_data_loader)
+    train_losses, valid_losses = fit(config['max_epochs'], model, loss_func, optimizer, train_data_loader, valid_data_loader, device)
     train_end_time = str(datetime.now())
     print('Finished Training')
 
--- a/3_train/src/train.py
+++ b/3_train/src/train.py
@@ -281,11 +281,11 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, device, verbose=False
     def verbose_print(*args):
         # Only print if verbose is true
         if verbose:
-            print(*args)
+            print(*args, flush=True)
 
     train_losses = []
     valid_losses = []
-    print('Epoch: train loss, validate loss')
+    print('Epoch: train loss, validate loss', flush=True)
     # Training loop
     for epoch in range(epochs):
         model.train()
@@ -315,7 +315,7 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, device, verbose=False
                 verbose_print(batch_loss, batch_count)
                 valid_loss += batch_loss * batch_count/n_valid
 
-        print(f'{epoch}: {train_loss}, {valid_loss}')
+        print(f'{epoch}: {datetime.now()} {train_loss}, {valid_loss}', flush=True)
         train_losses.append(train_loss)
         valid_losses.append(valid_loss)
     return train_losses, valid_losses
@@ -478,11 +478,11 @@ def main(npz_filepath, weights_filepath, metadata_filepath, run_id, model_id, co
         device = "cuda" 
     else:
         device = "cpu"
-    print(f"Using {device} device")
+    print(f"Using {device} device", flush=True)
 
-    print(f"Number of threads set by user: {torch.get_num_threads()}")
-    print(f"Number of GPUs: {torch.cuda.device_count()}")
-    print(f"Number of CPUs: {os.cpu_count()}")
+    print(f"Number of threads set by user: {torch.get_num_threads()}", flush=True)
+    print(f"Number of GPUs: {torch.cuda.device_count()}", flush=True)
+    print(f"Number of CPUs: {os.cpu_count()}", flush=True)
 
     # Create model
     model = get_model(n_depths,
@@ -505,7 +505,7 @@ def main(npz_filepath, weights_filepath, metadata_filepath, run_id, model_id, co
     train_start_time = str(datetime.now())
     train_losses, valid_losses = fit(config['max_epochs'], model, loss_func, optimizer, train_data_loader, valid_data_loader, device)
     train_end_time = str(datetime.now())
-    print('Finished Training')
+    print('Finished Training', flush=True)
 
     # Save model weights
     save_weights(model, weights_filepath, overwrite=True)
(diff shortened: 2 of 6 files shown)


sn_gpu.slurm
```diff
@@ -0,0 +1,27 @@
...
+
+run_id=initial1
+model_id=gpu_a
+log_dir=log/${run_id}_${model_id}
+mkdir -p ${log_dir}
+# From Jeff's blogpost: https://github.com/jsadler2/wdfn-blog/blob/ca63d4dc4bf6aeb6962a58953057a41c9db1ad8a/content/snakemake-ml-experiments.md
+# Also see snakemake profiles: https://snakemake.readthedocs.io/en/stable/executing/cli.html#profiles
+# srun snakemake --cluster "sbatch -A watertemp -t 1-23:59:59 -p gpu -N 1 -n 1 --job-name=${run_id}_${model_id} --gres=gpu:1 -e ${log_dir}/slurm-%j.out -o ${log_dir}/slurm-%j.out" -p -k --cores all --rerun-incomplete 3_train/out/mntoha/${run_id}/${model_id}_weights.pt 2>&1 | tee ${log_dir}/run.out
+snakemake --cluster "srun -A watertemp -t 1-23:59:59 -p gpu -N 1 -n 1 --job-name=${run_id}_${model_id} --gres=gpu:1 -e ${log_dir}/slurm-%j.out -o ${log_dir}/slurm-%j.out" --printshellcmds --keep-going --cores all --jobs 1 --rerun-incomplete 3_train/out/mntoha/${run_id}/${model_id}_weights.pt 2>&1 | tee ${log_dir}/run.out
+
```
hcorson-dosch, line 27: See my comments on the cpu batch script, above.

hcorson-dosch, line 27: See my comments on the cpu batch script, above.

sn_cpu.slurm
```diff
@@ -0,0 +1,26 @@
...
+
+## Run the main job
+
+run_id=test1
+model_id=cpu_a
+log_dir=log/${run_id}_${model_id}
+mkdir -p ${log_dir}
+# From Jeff's blogpost: https://github.com/jsadler2/wdfn-blog/blob/ca63d4dc4bf6aeb6962a58953057a41c9db1ad8a/content/snakemake-ml-experiments.md
+# Also see snakemake profiles: https://snakemake.readthedocs.io/en/stable/executing/cli.html#profiles
+snakemake --cluster "srun -A watertemp -t 1-23:59:59 -p cpu -N 1 -n 1 -c 16 --job-name=${run_id}_${model_id} -e ${log_dir}/slurm-%j.out -o ${log_dir}/slurm-%j.out" --printshellcmds --keep-going --cores all --jobs 1 --rerun-incomplete 3_train/out/mntoha/${run_id}/${model_id}_weights.pt 2>&1 | tee ${log_dir}/run.out
```
hcorson-dosch, line 25: I've not launched parallel tasks with slurm before, so I'm not sure that I can review this bit - you might want to run it by Jake or Jeff or Jeremy?

If it's running as expected as is, maybe this doesn't even need to be reviewed?

_______________________________

Here are my thoughts (keep that ^ caveat in mind):

When you launch this are you running `sh setup.sh` followed by `sbatch sn_cpu.slurm` (if running on the cpu partition), or `sbatch sn_gpu.slurm` (if running on the gpu portion)?

Here, you've set up a sbatch script in a `.slurm` file. The sbatch script requests an allocation, specifying 8 options (`-A`, `-p`, `-J`, `-t`, `-n`, `-c`, `-o`, and `-e`). Once it has that allocation, it submits a `Snakemake` `--cluster` command that includes a `srun` command that includes a _second_ allocation request, this time specifying 9 options (adding in the `-N` option to specify the number of nodes).

I'm a bit unsure about what's happening here with the double allocation request. It seems like the use of `srun` is right, per [the docs](https://hpcportal.cr.usgs.gov/training/courses/Intro_to_HPC/10_slurm.html#priority-reservations):
> **Execution with srun**
`srun [OPTIONS(0)...] [ : [OPTIONS(N)...]] executable(0) [args(0)...]`
> * Runs a parallel job on cluster managed by Slurm. If necessary, srun will first create a resource allocation in which to run the parallel job.
> * Within an allocation, srun is used to determine what runs on any given subset of resources available
> * Multiple srun commands can be used sequentially or concurrently within a job
> * Failing to use srun within a job allocation will result in the executables being run only on a single node, with an undetermined set of resources. This leads to resource contention and generally bad behavior

...but I'm confused about what that allocation request within the `srun` command is doing. My initial thinking was, given that the `srun` allocation request is identical to the `sbatch` allocation request, is it needed here? But would guess there's a reason you included a second allocation request, and suspect I'm just missing some nuance of slurm. When you run this, what job names, etc. do you see in `squeue`? Are they named `snakemake`, or `test1_cpu_a`? Or is there a single `snakemake` job and then many `test1_cpu_a` jobs?  Also I see that you don't specify the number of nodes in your `sbatch` request, but you do in  your `srun` request. Since the default is 1 task per node and you're only requesting `-n 1` does it not matter?

_To help me think this through, I checked out these other sbatch uses:_
  * In Jeff's blog post, to run tasks in parallel, he submitted a [sbatch command](https://github.com/jsadler2/wdfn-blog/blob/snakemake-dl-experiments/content/snakemake-ml-experiments.md#executing-snakemake-on-many-cores) _within_ a `Snakemake` command

  * For `lake-temperature-neural-networks`, to run tasks in serial, we ran an R script in a [`slurm.sbatch` file](https://code.usgs.gov/hcorson-dosch/lake-temperature-neural-networks/-/blob/master/2_model/src/slurm.sbatch) by [calling `sbatch 2_model/src/slurm.sbatch` in the terminal](https://code.usgs.gov/hcorson-dosch/lake-temperature-neural-networks/-/blob/master/2_model/src/model_tasks.R#L139-142).

hcorson-dosch, line 25: I've not launched parallel tasks with slurm before, so I'm not sure that I can review this bit - you might want to run it by Jake or Jeff or Jeremy?

If it's running as expected as is, maybe this doesn't even need to be reviewed?

_______________________________

Here are my thoughts (keep that ^ caveat in mind):

When you launch this are you running `sh setup.sh` followed by `sbatch sn_cpu.slurm` (if running on the cpu partition), or `sbatch sn_gpu.slurm` (if running on the gpu portion)?

Here, you've set up a sbatch script in a `.slurm` file. The sbatch script requests an allocation, specifying 8 options (`-A`, `-p`, `-J`, `-t`, `-n`, `-c`, `-o`, and `-e`). Once it has that allocation, it submits a `Snakemake` `--cluster` command that includes a `srun` command that includes a _second_ allocation request, this time specifying 9 options (adding in the `-N` option to specify the number of nodes).

I'm a bit unsure about what's happening here with the double allocation request. It seems like the use of `srun` is right, per [the docs](https://hpcportal.cr.usgs.gov/training/courses/Intro_to_HPC/10_slurm.html#priority-reservations):
> **Execution with srun**
`srun [OPTIONS(0)...] [ : [OPTIONS(N)...]] executable(0) [args(0)...]`
> * Runs a parallel job on cluster managed by Slurm. If necessary, srun will first create a resource allocation in which to run the parallel job.
> * Within an allocation, srun is used to determine what runs on any given subset of resources available
> * Multiple srun commands can be used sequentially or concurrently within a job
> * Failing to use srun within a job allocation will result in the executables being run only on a single node, with an undetermined set of resources. This leads to resource contention and generally bad behavior

...but I'm confused about what that allocation request within the `srun` command is doing. My initial thinking was, given that the `srun` allocation request is identical to the `sbatch` allocation request, is it needed here? But would guess there's a reason you included a second allocation request, and suspect I'm just missing some nuance of slurm. When you run this, what job names, etc. do you see in `squeue`? Are they named `snakemake`, or `test1_cpu_a`? Or is there a single `snakemake` job and then many `test1_cpu_a` jobs?  Also I see that you don't specify the number of nodes in your `sbatch` request, but you do in  your `srun` request. Since the default is 1 task per node and you're only requesting `-n 1` does it not matter?

_To help me think this through, I checked out these other sbatch uses:_
  * In Jeff's blog post, to run tasks in parallel, he submitted a [sbatch command](https://github.com/jsadler2/wdfn-blog/blob/snakemake-dl-experiments/content/snakemake-ml-experiments.md#executing-snakemake-on-many-cores) _within_ a `Snakemake` command

  * For `lake-temperature-neural-networks`, to run tasks in serial, we ran an R script in a [`slurm.sbatch` file](https://code.usgs.gov/hcorson-dosch/lake-temperature-neural-networks/-/blob/master/2_model/src/slurm.sbatch) by [calling `sbatch 2_model/src/slurm.sbatch` in the terminal](https://code.usgs.gov/hcorson-dosch/lake-temperature-neural-networks/-/blob/master/2_model/src/model_tasks.R#L139-142).

hcorson-dosch (approved):
Hi Andy -- I apologize, I don't think I can offer much of a review here. I've left my (somewhat confused) thoughts about the slurm script in a comment. Feel free to respond/answer my Qs if you think doing so would help refine these scripts, but if everything is running as expected and you'd like to move ahead, then I think you can ignore my Qs. If the slurm commands are _not_ running as expected, I think it would be worth reaching out to Jake/Jeff/Jeremy/Jesse, as I've not submitted parallel tasks using `sbatch` or `snakemake` before.

3_train/src/train.py
```diff
@@ -286,18 +281,20 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, verbose=False):
     def verbose_print(*args):
         # Only print if verbose is true
         if verbose:
-            print(*args)
+            print(*args, flush=True)
```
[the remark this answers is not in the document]
AndyMcAliley: I needed it to get updates as the model ran. Otherwise, I got no feedback until everything had finished, so there was no way to tell how close it was to finishing!

3_train/src/train.py
```diff
@@ -286,18 +281,20 @@ def fit(epochs, model, loss_func, opt, train_dl, valid_dl, verbose=False):
...
+    print('Epoch: train loss, validate loss', flush=True)
     # Training loop
     for epoch in range(epochs):
         model.train()
         train_loss = 0.0
         # Data is ordered as [dynamic inputs, static inputs, labels]
         verbose_print('Train loss by batch')
         for x_d, x_s, y in train_dl:
+            # Transfer to device
+            x_d, x_s, y = x_d.to(device), x_s.to(device), y.to(device)
```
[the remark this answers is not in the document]
AndyMcAliley: New to me, too. It's a command that is specific to PyTorch. But, without it, PyTorch threw an error about tensors being on a different device than the model.

.gitignore
```diff
@@ -166,6 +168,7 @@ dmypy.json
 2_process/tmp
 2_process/out
 3_train/out
+log
```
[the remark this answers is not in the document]
AndyMcAliley: Nice post - thanks! I'll change that line along with the ones above it - they should all pertain to folders in the root of the repository only.

sn_cpu.slurm
AndyMcAliley: These are good questions, and while the thing runs now, I don't think that what I've got is perfectly correct. I am using the slurm script in the way you described. Here's how I'm using it:

1. run `./setup.sh` (or, rather, the individual commands in it, since `conda activate` doesn't work inside the script)
2. `sbatch sn_gpu.slurm` or `sbatch sn_cpu.slurm`. 

Here's what I'm hoping it does:
1. Acquire an allocation on gpu/cpu as needed
2. Run the snakemake command on that allocation.
3. Train the model(s) _on that same allocation_ (not sure if this is actually what's happening, but I think it is. Note that I've only tried training one model at a time so far).
4. Use only one node per model, regardless of how many I have allocated to me (no benefit in training one model using multiple nodes).

So, I'm hoping that each `srun` command trains a model on the current allocation, instead of requesting a new allocation. When I run `squeue` I see only one job named snakemake while the model is training. A good sign!

If I use `sbatch` within the snakemake command, like Jeff did, then I think snakemake will make a new allocation request for every model. If the queue time on Tallgrass is long, then the allocation that Snakemake itself runs on might time out before these new allocations start. Maybe one could run snakemake on the head node, but that seems like a bad idea?

If I don't include any --cluster command at all, I'm not sure what would happen. Given the admonition in the `srun` docs that you shared about what happens when you don't use `srun` in an allocation, that seems like a bad idea, too.

I didn't specify the number of nodes in `sbatch` but since `n_tasks=1` and `cpus_per_task=1`, I think I should only get one cpu on one node allocated. I'm thinking of each model that gets trained as one task. I expect to modify `n_tasks` as needed in order to do grid searches for hyperparameter tuning.

I think I should remove `-c 16` from the `srun` command, and let that value be determined by `cpus_per_task` from `sbatch`. I wonder if some of the other flags can be removed, too (`-A`, `-t`, `-p`).

hcorson-dosch: > Here's what I'm hoping it does:
> 1. Acquire an allocation on gpu/cpu as needed
> 2. Run the snakemake command on that allocation.
> 3. Train the model(s) on that same allocation (not sure if this is actually what's happening, but I think it is. Note that I've > only tried training one model at a time so far).
> 4. Use only one node per model, regardless of how many I have allocated to me (no benefit in training one model > using multiple nodes).

Thanks for outlining that ^ -- That all sounds good, and it seems like your set up will achieve that, with the possible exception of 3 based on the `srun` options. But I think if you try out training >1 model it _should_ become apparent whether or not the same allocation is being used to submit the `snakemake` command and train the models. I agree that the fact that you only see one job named `snakemake` while a single model is training is a good sign. I _think_ that reflects `srun` recognizing that it doesn't need to create an allocation, since one already exists? But I am curious what job names would be displayed in `squeue` if you trained >1 models. Also, in addition to testing w/ >1 training run I think you could also try dropping some of the options (or all but -J -e and -o) after `srun`, as you suggest.

hcorson-dosch (approved):
Hmm Github is being strange for me on this PR review... do you also see my comments x2?

AndyMcAliley:
Yes, @hcorson-dosch, I'm seeing 2x comments on both this pull request and #32

sn_cpu.slurm
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+#
+#SBATCH --account=watertemp
+#SBATCH --partition=cpu
+#SBATCH --job-name=snakemake
+#SBATCH --time=1-23:59:59
+#SBATCH --ntasks=1
```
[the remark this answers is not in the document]
AndyMcAliley: I've added `--nodes` here in the `sbatch` directives.

sn_cpu.slurm
AndyMcAliley: I tried dropping the options `-A`, `-t`, `-p` and `-c`, and everything ran the same.

AndyMcAliley: I tried training multiple models as well. The same allocation was used, BUT the models were trained in serial. Uh-oh. It seems that `srun` needs an ampersand after its command in order to run multiple `srun` steps in parallel. Snakemake doesn't make this easy because the input to `--cluster` is basically a prefix, not a suffix. There are two options that I see:

1. Change `srun` to `sbatch`, like Jeff does. That will queue each model up for a new allocation, rather than using the current allocation.
2. In `3_train.smk`, in the rule `train_model`, change `script` to `shell` and hardcode a `srun` command there with an ampersand at the end of it. Then, in `3_train/src/train.py`, change the script part of the file so that it no longer expects snakemake to be an object, and so that it uses command line arguments to determine output file paths, config, etc. instead.

Since option 1 is easier to implement, I'm going that route for now.

Issue #31, opened 2022-05-05
Split validation data by lake


AndyMcAliley:
Test data is split by lake, but validation data is not. This can lead to overfitting, since the validation data used to determine early stopping criteria is taken from the same lakes as the training data. This split should happen during `2_process` at the same time that the test set is formed, so we can remove the training/validation splitting that happens in `3_train`.

AndyMcAliley:
Closed by #37

PR #37, opened 2022-06-16
Split validation set during 2_process

AndyMcAliley:
This PR splits validation data at the same point in the workflow and in the same manner as the testing data are split. The data are split by lake rather than by sequence. Splitting by lake means that the lakes used for validation data are always different than the lakes used for training data (and different than the lakes used for testing data). That way, validation metrics will better represent the model's performance on new lakes that aren't in the training set. So, early stopping should be even more effective at avoiding overfitting to the training data.

# How to run the code
The rule `create_training_data` creates training, validation, and testing sets. So, to run this part of the pipeline:
```
```
That should make `train.npz`, `valid.npz`, and `test.npz`.

# How to review this PR

## Level of review requested

The code seems to work - no lake site IDs appear in more than one split.

2. Have any errors been introduced into the pipeline?

## Where in the code to focus

Anything that's been changed is fair game, but the majority of the changes are in `training_data.py`.

- Save more metadata alongside train.npz, validate.npz, and test.npz for use during model evaluation

### DIFF
--- a/2_process.smk
+++ b/2_process.smk
@@ -456,8 +456,9 @@ rule create_training_data:
         sequences_summary_file = "2_process/out/{data_source}/sequences/{data_source}_sequences_summary.csv"
     output:
         train_file = "2_process/out/{data_source}/train.npz",
+        valid_file = "2_process/out/{data_source}/valid.npz",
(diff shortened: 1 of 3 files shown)


AndyMcAliley:
Closes #31.

PR #38, opened 2022-06-23
Save training metadata

AndyMcAliley:
This PR ensures that metadata are saved alongside the sequences in the training, validation, and testing sets.

Previously, when data were organized into `train.npz`, `valid.npz`, and `test.npz` as equal-length sequences, information about **WHEN** those sequences start or **WHICH LAKE** they belong to was not saved, because that information isn't relevant to training the neural network. However, that information is relevant to evaluating the neural network. Therefore, in this PR those metadata are added to the npz files. That way, `test.npz` can be used to examine model performance during different seasons and in specific lakes or sets of lakes.

The metadata are saved in the three npz files as as additional arrays with the same length as the number of sequences. That is, there is one start date and one lake ID per sequence.

# How to run the code
The rule `create_training_data` creates training, validation, and testing sets. So, to run this part of the pipeline:
```
```
That should make `train.npz`, `valid.npz`, and `test.npz`, now with additional files `start_dates` and `site_ids`.

# How to review this PR

## Level of review requested

The code seems to work - data appear in the npzs where expected.

2. Have any errors been introduced?

## Where in the code to focus

Anything that's been changed is fair game, but the majority of the changes are in `lake_sequences.py` and `training_data.py`.

- Changes to 3_train functions that accommodate these updates

### DIFF
--- a/2_process.smk
+++ b/2_process.smk
@@ -450,7 +450,7 @@ rule process_sequences:
 rule create_training_data:
     input:
         lambda wildcards: get_lake_sequence_files(
-            '2_process/out/{}/sequences/sequences_{}.npy',
+            '2_process/out/{}/sequences/sequences_{}.npz',
             wildcards.data_source
         ),
         sequences_summary_file = "2_process/out/{data_source}/sequences/{data_source}_sequences_summary.csv"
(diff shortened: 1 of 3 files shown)


2_process.smk
```diff
@@ -424,7 +424,7 @@ def save_sequences_summary(lake_sequence_files_input, summary_file):
     sequence_counts = []
     for sequences_file in lake_sequence_files_input:
         # Sequence files have shape (# sequences, sequence length, # depths + # features)
```
hcorson-dosch, line 426:
```suggestion
        # The `lake_sequences` elements of the sequence files have shape (# sequences, sequence length, # depths + # features)
```

hcorson-dosch (approved):
Looks good, Andy. It's nice to have this metadata captured and stored with the sequences

Issue #35, opened 2022-06-06
No automatic early stopping


AndyMcAliley:
During training, we want to save the set of model weights that corresponds to the training epoch with the lowest validation loss. Also, we want to stop training once we're confident that the validation loss isn't going to decrease more. Loss values tend to bounce around from epoch to epoch, so we'll want to give it a few epochs rather than stop after the first epoch when validation loss increases. So, to implement early stopping,

1. Save model weights for the lowest validation loss so far.
2. Stop training if validation loss hasn't decreased for N epochs.

AndyMcAliley:
Closed by #40

PR #40, opened 2022-07-27
Early stopping

AndyMcAliley:
This PR implements early stopping with a user-defined patience. With early stopping implemented, training will continue until either

1. `early_stopping_patience` training epochs pass without a decrease in validation loss, or
2. `max_epochs` is reached.

These variables are defined in `3_train/in/{data_source}/{run_id}.yaml`. Model weights for the epoch with the lowest validation loss (not necessarily the final training epoch) are saved.

# How to run the code
These changes all concern the rule `train_model` in `3_train.smk`. So, to run this part of the pipeline, first copy `3_train/in/example_config.yaml` to `3_train/in/model_prep/1.yaml`. Then:
```
snakemake --cores all 3_train/out/model_prep/1/a_weights.pt
```
That should make `3_train/out/model_prep/1/1_process.yaml`, `3_train/out/model_prep/1/a_metadata.npz`, `3_train/out/model_prep/1/a_weights.pt`, and `3_train/out/model_prep/1/1_train.yaml`.

# How to review this PR

## Level of review requested

1. Does the implementation of early stopping look correct?
2. Should any related code additions or tweaks be made?
3. Have any errors been introduced?
4. Are the comments clear?

## Where in the code to focus

Anything that's been changed is fair game!

## Issues that are slated for upcoming PRs (so don't worry about them yet)

### DIFF
--- a/3_train/src/train.py
+++ b/3_train/src/train.py
@@ -10,7 +10,7 @@ from torch.utils.data import Dataset, DataLoader, random_split
 from torch.optim import Adam
 
 import sys
-sys.path.insert(1, '3_model/src')
+sys.path.insert(1, '3_train/src')
 from models import Model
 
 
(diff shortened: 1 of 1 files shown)


3_train/src/train.py
```diff
@@ -320,7 +329,22 @@ def verbose_print(*args):
...
+            saved_epoch = epoch
+            save_weights(model, weights_filepath, overwrite=True)
+        # Otherwise, add to count of high validation loss models and check for early stopping condition
+        else:
+            num_epochs_without_improvement += 1
+            # An early_stopping_patience of -1 means early stopping should not happen
+            if early_stopping_patience != -1:
+                if num_epochs_without_improvement >= early_stopping_patience:
+                    print(f'Stopping at epoch {epoch} after {num_epochs_without_improvement} epochs without improvement', flush=True)
+                    break
```
jdiaz4302, line 345: If I'm reading this right, the psuedocode would be:

```
*(only) if lowest validation loss achieved:* # still happens when early stopping is disabled
    *save weights*
if lowest validation loss not achieved:
    track patience
    *if early stopping is enabled:*
        if patience exceeded:
            stop
```
I believe the asterick'd line orders are problematic because if you intended early stopping to be disabled, the code will still be only saving weights from the best epoch (which could be epoch 500 of 1000), thus still giving you early stopped model weights but having done all the non-early stopped epochs.

If you intend to keep the functionality of disabling early stopping, the `if early_stopping_patience != -1` (which indicates if early stopping is disabled/enabled) should be repositioned or duplicated to decide whether we save models weights associated with the best epoch for validation loss or every epoch regardless of better/worse validation loss.

jdiaz4302 (commented):
It seems that the goal is to add early stopping as an option. I think there is a small bug in the code's implementation when disabling early stopping via `fit(..., early_stopping_patience = -1)` because the function always only saves the weights for the lowest validation loss epoch without consideration of `early_stopping_patience = -1`

3_train/src/train.py
AndyMcAliley: Nice catch! You're right. The bug was in my brain - I intentionally coded it this way, not thinking of the possibility that we'd want to save the weights for a model that doesn't have the lowest validation loss. I'll make the change.

Note that weights aren't saved for every epoch - only one set of weights is being saved per training run, for the sake of performance and memory savings. In the future we can add an option to save weights at each epoch if/when we need it.

jdiaz4302:
The changes look good to me

jdiaz4302 (approved):

AndyMcAliley:
Closes #35

PR #47, opened 2022-09-01
Interpolate predictions

### LINKED ISSUE #46: Interpolate predictions to observation depths
The direct outputs of the ML model are predicted temperatures at many fixed depths per day. Let's interpolate those predictions to the depths of observations for the sake of model evaluation, and for consistency with lake-temperature-process-models (see [this issue](https://github.com/USGS-R/lake-temperature-process-models/issues/59) for discussion and details).

Do this for the train, validation, and test sets.

AndyMcAliley:
Interpolate predicted temperatures from fixed depths to depths of observed temperatures.

The LSTM predicts temperatures at a set of fixed depths at every time step (day). Here we interpolate those predictions to the depths of every observation in a training, validation, or test set for the sake of model evaluation.

The trickiest part is to match each observation to the right prediction sequence. Training, validation, and testing sequences are formed as follows. Every sequence is a fixed number of days (set to 400 days right now). The first part of the sequence is "spin-up" or "burn-in" time (set to 100 days right now). Only predictions after spin-up are used for model evaluation. Sequences can overlap (set to 200 days right now). Only sequences that have an observation occur after spin-up are included in a dataset. Every observation for which we have NLDAS driver data has at least one corresponding sequence. Because of sequence overlap, some observations have more than one corresponding sequence.

When it comes to finding the right sequence for each observation, performance matters. My original implementation was a set of nested for loops over all observations and all sequences. It was on track to take 20 hours for the validation set. The current implementation takes 2-3 minutes. I'm using a combination of batch processing, [broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html), [np.nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html), and [np.where](https://numpy.org/doc/stable/reference/generated/numpy.where.html) to take advantage of vectorized operations.

The question of how to evaluate model performance was first raised in lake-temperature-process-models, [here](https://github.com/USGS-R/lake-temperature-process-models/issues/59). With this PR we can evaluate models in the same way as lake-temperature-process-models.

# How to run the code
The rule `interpolate_predictions` is the new addition. So, to run this part of the pipeline:
```
snakemake -c1 -s 4_evaluate.smk 4_evaluate/out/model_prep/initial/cpu_a/interpolated_predictions_valid.csv
```
That should make `interpolated_predictions_valid.csv` with the column `predicted_temperature_obs_depth` that contains the interpolated predictions.

> NOTE: Running the command above requires a compiled dataset and a trained model each of which takes several hours to obtain, plus a set of observed temperatures. If you want, copy the following files from Tallgrass, preserving their directory structure relative to the repo root directory:
> - 2_process/out/model_prep/valid.npz
> - 2_process/tmp/model_prep/temperature_observations_interpolated.csv
> - 3_train/out/model_prep/initial/cpu_a_metadata.npz
> - 3_train/out/model_prep/initial/cpu_a_weights.npz

# How to review this PR

2. Have any errors been introduced into the pipeline?

Closes #46.

### DIFF
--- a/4_evaluate/src/interpolate_predictions.py
+++ b/4_evaluate/src/interpolate_predictions.py
@@ -66,10 +66,6 @@ def get_interpolated_predictions(predictions_filepath,
     start_dates = dataset_npz['start_dates']
     sequence_length = dataset_npz['sequence_length']
     spinup_time = dataset_npz['spinup_time']
-    sequence_offset = dataset_npz['sequence_offset']
-    depths_all = dataset_npz['depths_all']
-    dynamic_features_all = dataset_npz['dynamic_features_all']
-    static_features_all = dataset_npz['static_features_all']
 
     # Load metadata to get depths_use
     metadata_npz = np.load(metadata_filepath)
@@ -145,8 +141,8 @@ def get_interpolated_predictions(predictions_filepath,
     clipped_obs_depths = np.clip(obs_depths, min(depths_use), max(depths_use))
     # Pad depths with one value that is always below obs,
     # for convenience in interpolation algorithm below
-    pad_value = max(np.r_[depths_all, clipped_obs_depths] + 1)
-    padded_pred_depths = np.r_[depths_all, pad_value]
+    pad_value = max(np.r_[depths_use, clipped_obs_depths] + 1)
+    padded_pred_depths = np.r_[depths_use, pad_value]
 
     # Get index and value of next largest depth in list of depths
     def enumerate_depth_below(depth):
@@ -156,7 +152,7 @@ def get_interpolated_predictions(predictions_filepath,
     # Unpack list of two-element tuples into two arrays
     depth_below_index, depth_below = (np.array(tp) for tp in zip(*depth_below_tuples))
     depth_above_index = depth_below_index - 1
-    depth_above = depths_all[depth_above_index]
+    depth_above = depths_use[depth_above_index]
 
     # Compute interpolation factors
     above_factors = (depth_below - clipped_obs_depths)/(depth_below - depth_above)
--- a/4_evaluate/src/interpolate_predictions.py
+++ b/4_evaluate/src/interpolate_predictions.py
@@ -211,7 +211,7 @@ def get_interpolated_predictions(predictions_filepath,
 
     # Add column named pred to obs and populate with NaN
     obs_preds = obs.assign(pred=np.NaN)
-    obs_preds.loc[use_obs, 'pred'] = predictions
+    obs_preds.loc[use_obs, 'predicted_temperature_obs_depth'] = predictions
     obs_preds.to_csv(interpolated_predictions_filepath, index=False)
 
 
--- a/4_evaluate/src/interpolate_predictions.py
+++ b/4_evaluate/src/interpolate_predictions.py
@@ -50,7 +50,8 @@ def get_interpolated_predictions(predictions_filepath,
     :param predictions_filepath: Path to unnormalized predictions for one
         dataset (train, validation, or test)
     :param metadata_filepath: Path to model training metadata npz for one dataset
-    :param dataset_filepath: Path to input for one dataset
+    :param dataset_filepath: Path to one dataset's npz containing features and
+        observations
     :param obs_file: File with all temperature observations
     :param interpolated_predictions_filepath: csv file to save interpolated
         predictions to
@@ -76,7 +77,6 @@ def get_interpolated_predictions(predictions_filepath,
     obs_all = pd.read_csv(obs_file, index_col=0, parse_dates=['date'])
     # Whittle down to lakes in this dataset
     obs = obs_all[obs_all.site_id.isin(site_ids)]
-    obs_all.site_id.unique().shape, obs.site_id.unique().shape, np.unique(site_ids).shape
 
     # Prepare to find the indices of the predictions that are needed to
     # interpolate temperatures to observation depths, for every observation in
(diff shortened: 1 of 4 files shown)


4_evaluate/src/interpolate_predictions.py
```diff
@@ -0,0 +1,225 @@
...
+                                 obs_file,
+                                 interpolated_predictions_filepath,
+                                 batch_size=5000):
+    """
+    Interpolate ML output to observation depths.
+
+    :param predictions_filepath: Path to unnormalized predictions for one
+        dataset (train, validation, or test)
+    :param metadata_filepath: Path to model training metadata npz for one dataset
+    :param dataset_filepath: Path to input for one dataset
```
jdiaz4302, line 53: "path to input for one dataset" is vague/not as informative as the other lines of documentation

4_evaluate/src/interpolate_predictions.py
```diff
@@ -0,0 +1,225 @@
...
+    # Load metadata to get depths_use
+    metadata_npz = np.load(metadata_filepath)
+    depths_use = metadata_npz['depths_use']
+
+    # Now, we'll need obs for each of those lakes.
+    # Read temperature observations
+    obs_all = pd.read_csv(obs_file, index_col=0, parse_dates=['date'])
+    # Whittle down to lakes in this dataset
+    obs = obs_all[obs_all.site_id.isin(site_ids)]
+    obs_all.site_id.unique().shape, obs.site_id.unique().shape, np.unique(site_ids).shape
```
jdiaz4302, line 79: Does this actually print from snakemake / is it useful or is this a development relic?

If it is a development aid, perhaps an `assert(shape1 == shape2)` (like those you have later) would be a useful test to have here instead.

4_evaluate/src/interpolate_predictions.py
```diff
@@ -0,0 +1,225 @@
...
+            # Select the sequence with the earliest start date
+            batch_start_dates = pred_starts_days[idx_sequence]
+            # Create a 2D array with the matches along rows and the observations in
+            # this batch along columns. Values are start dates where a match
+            # corresponds to an observation in this batch, and NaN otherwise.
+            start_date_matrix = np.where(
+                idx_obs[:, np.newaxis] == np.arange(this_batch_size),
+                batch_start_dates[:, np.newaxis],
+                np.NaN
+            )
```
jdiaz4302, line 144: It seems to me that `idx_obs` is expected to be a shape less than `this_batch_size` due to the triple condition subsetting/index identification that happened earlier (Line 112). Is that intuition right? It seems that this would raise an error if that were the case.

For example, if the `batch_size` is 5 and `idx_obs` identifies the first observation as failing to meet the earlier conditions, we would get `idx_obs = [1, 2, 3, 4]` (right?) which would then error out when compared to `np.arange(5)`

4_evaluate/src/interpolate_predictions.py
```diff
@@ -0,0 +1,225 @@
...
+    # Good. That means that the depth_below_index doesn't need to be used when it's one greater than max
+    # because those values get multiplied by zero
+    # Set those indices to the max index instead
+    depth_below_index[too_low] = len(depths_use) - 1
+
+    # Let's now skip any obs with indices equal to -1
+    # TODO: Consider dropping observations that are deeper than deepest depth_use
+    use_obs = ((sequence_indices >= 0) & (date_indices >= 0))
+
+    # Now the tricky part (like it hasn't been tricky so far!):
```
jdiaz4302, line 199: Yes 😅. This looks like it was a lot of work that is tricky to think-up from scratch

jdiaz4302 (commented):
_🤦🏼‍♂️ I still don't have this pipeline runnable because the large files failed to download off Tallgrass a couple times and I moved on to other things, but I will give that a go again_

Despite not running the pipeline, I mocked up a few minimal example tests of a lot of this functionality (e.g., the triple-condition index identification, the depth matching, the interpolation weight determination). All those minimal tests behaved as expected and the flow of operations  makes sense and seems efficient,

Specifically, I had one misunderstanding or concern regarding the `state_date_matrix` that I'd like addressed before approval, but other than looks great and like hard work

4_evaluate/src/interpolate_predictions.py
AndyMcAliley: Agreed - reworded

4_evaluate/src/interpolate_predictions.py
AndyMcAliley: Whoops! Yes, this is a pure dev relic. I think I'll leave out asserts for now

4_evaluate/src/interpolate_predictions.py
AndyMcAliley: Your intuition is right: the length of `idx_obs` is generally not `this_batch_size`. `idx_obs` can be a different shape and this will work thanks to the broadcasting that's happening. If it were `idx_obs == np.arange(this_batch_size)` then it would error. To rephrase from the [numpy docs](https://numpy.org/doc/stable/user/basics.broadcasting.html#broadcastable-arrays):
> Here the `newaxis` index operator inserts a new axis into `idx_obs`, making it a two-dimensional nx1 array. Comparing the nx1 array with `np_arange(this_batch_size)`, which has shape `(this_batch_size,)`, yields a nx`this_batch_size` array.

It reminds me of an [outer product](https://en.wikipedia.org/wiki/Outer_product) in that one vector gets transposed and the result is a matrix where every element (i, j) is a function of the ith element of one vector and the jth element of the other.

4_evaluate/src/interpolate_predictions.py
AndyMcAliley: Haha, maybe too tricky... but it's a working solution!

4_evaluate/src/interpolate_predictions.py
jdiaz4302: Ah, okay. I see, and this checks out. 

broadcastingcheck

jdiaz4302 (approved):

4_evaluate/src/interpolate_predictions.py
AndyMcAliley: Nice explanatory demo!

PR #48, opened 2022-09-09
Plot metrics

AndyMcAliley:
This PR adds the phase `5_visualize`. It allows RMS error and bias to be plotted by lake, by depth, or by day of year.

The plots rely on the interpolated predictions csv. Each row of that csv has both predictions and observations, which is convenient for computing residuals and plotting.

## How to run the code

- Install `ltls` conda environment from `environment_local.yaml` because additional plotting packages have been added.
`mamba` is much faster:
```bash
mamba env create -f environment_local.yaml
```
but `conda` should work
```bash
conda env create -f environment_local.yaml
```
- Create all validation plots.
```bash
snakemake -c all -r -s 5_visualize.smk 5_visualize/out/model_prep/initial/cpu_a/plot_all_valid.dummy
```

NOTE: The pipeline commands above require a compiled dataset and a trained model each of which takes several hours to obtain, plus a set of observed temperatures and metadata. If you want, copy the following files from Tallgrass, preserving their directory structure relative to the repo root directory:
- 2_process/out/model_prep/train.npz
- 2_process/out/model_prep/valid.npz
- 2_process/tmp/model_prep/temperature_observations_interpolated.csv
- 2_process/tmp/model_prep/lake_metadata_augmented.csv
- 3_train/out/model_prep/initial/cpu_a_metadata.npz
- 3_train/out/model_prep/initial/cpu_a_weights.npz

The prediction/interpolation takes 20-30 minutes. To skip that as well, you should only need the following files:
- 2_process/tmp/model_prep/lake_metadata_augmented.csv
- 4_evaluate/out/model_prep/initial/cpu_a/interpolated_predictions_valid.csv
- 4_evaluate/out/model_prep/initial/cpu_a/interpolated_predictions_train.csv

## How to review this PR

- Are there any obvious errors?
- Suggestions for additional plots for metrics are welcome

### Issues that will be addressed in upcoming PRs (so don't worry about them yet)

- Change to a different plotting library - probably `seaborn`


5_visualize/src/plot_metrics.py
```diff
@@ -0,0 +1,229 @@
...
+                p = df_plot.hvplot.line() * zero_line
+            else:
+                p = df_plot.hvplot.line()
+        elif plot_by == 'doy':
+            if metric == 'bias':
+                p = df_plot.hvplot.line() * zero_line
+            else:
+                p = df_plot.hvplot.line()
+        else:
+            raise ValueError(f'Evaluating metrics by {plot_by} is not supported')
```
jdiaz4302, line 166: I'm confused about the conditional statements (lines 155 and 160) based on `plot_by` equaling `depth` or `doy`. It seems that they're identical. Would it be equivalent and more compact to having something like this instead:

```
if plot_by not in ['depth', 'doy']:
    raise ValueError(f'Evaluating metrics by {plot_by} is not supported')
else:
    if metric == 'bias':
        zero_line = hv.HLine(0).opts(color='black', level='underlay')
        p = df_plot.hvplot.line() * zero_line
    else:
        p = df_plot.hvplot.line()
```

If identical, the arrangement of lines is trivial, but, especially being unfamiliar with holoviews, the current arrangement makes me wonder if holoviews is accessing the `plot_by` object in some hidden/unclear way.

jdiaz4302 (approved):

jdiaz4302:
No obvious errors stood out to me. The use and flow of operations looked good and the final plots do not look suspicious. 

The bias/rmse for interpolated depth has that obvious spike at around 90 meters that kind of begs context. I assume that may be a single lake that is only responsible for observations that depth? It may be worth seeing something like decile ticks (i.e., 10th, 20th, 30th, etc... percentiles) along the x-axis to get an idea of data density/sparsity

Some other potentially interesting dimensions to view error/bias along might include elevation and lake area.

5_visualize/src/plot_metrics.py
AndyMcAliley: Yes, you're right. That would be equivalent and more concise. There's no hidden use of `plot_by`. I think I'll leave it as it is for now, though, because I'll likely want the code arranged like this if/when I change to another plotting library. The plot axes will need to be labeled manually based on which variable we're plotting by.

PR #56, opened 2022-10-13
Write metrics

AndyMcAliley:
Write metrics evaluated on the training, validation, and testing sets out to csv files. Metrics are evaluated for one model run at a time. 

Residuals are computed by this sequence of steps.

1. Predict temperatures at a set of fixed depths using the model.
2. Interpolate predictions from fixed depths to actual observation depths.
3. Subtract interpolated predicted temperatures from observed temperatures and write them to a file.

Here's a file graph for computing metrics over the training set. This is only the part of the file graph that comes after training. It matches the three steps above.

metrics_filegraph

This PR is only concerned with step 3, and even then, only with the writing part. Computing residuals is achieved using an existing function in the codebase, `5_evaluate/src/plot_metrics.py:load_predictions`.

# How to run the code

`snakemake -s 4_evaluate.smk -c1 4_evaluate/out/model_prep/initial_ealstm/local_a/metrics_test.csv`
This code has been run locally and on Tallgrass with success.

# How to review this PR

I'm only looking for a quick review. Are there any obvious errors? I'm getting the codebase in order anticipating the data release, and this utility code has been sitting around in local repos for a while now.


4_evaluate/src/write_metrics.py
```diff
@@ -0,0 +1,50 @@
...
+    """
+    Compute metrics for a dataset, including overall RMSE, median and mean RMSE
+    by lake, the total number of observations, and the number of lakes.
+
+    :param predictions_filepath: Path to csv with model predictions
+    :returns: DataFrame of metrics computed for the predictions
+
+    """
+    preds = load_predictions(predictions_filepath)
+    rmse_lake = preds.groupby('site_id')['residual'].aggregate(rms).rename('LSTM_rmse')
```
jdiaz4302, line 23: Is the `.rename('LSTM_rmse')` doing anything?

It seems that that is more of a `pandas`-column-renaming thing, but `rmse_lake` is being used more-or-less as an array which doesn't need a name.

jdiaz4302 (approved):
Looks good - simple and straightforward.

4_evaluate/src/write_metrics.py
AndyMcAliley: You're right - I think that's a leftover from code that was using the `pandas` column names directly for plotting.
```suggestion
    rmse_lake = preds.groupby('site_id')['residual'].aggregate(rms)
```

AndyMcAliley:
Thanks!

